PR #47, opened 2018-11-07
Init OData connector

vdestraitt:


toucan_connectors/__init__.py
```diff
@@ -30,6 +30,8 @@
     from .mysql.mysql_connector import MySQLConnector
 with suppress(ImportError):
     from .oracle_sql.oracle_sql_connector import OracleSQLConnector
+with suppress(ImportError):
+    from .odata.odata_connector import ODataConnector
```
adimascio: Nitpicking here, sorry… but it looks like imports in `toucan_connectors/__init__` are sorted alphabetically. `odata` should therefore probably stand before `oracle`

tests/odata/test_odata.py
```diff
@@ -0,0 +1,66 @@
...
+        password='xxx')
+    data_source = ODataDataSource(
+        domain='test',
+        name='test',
+        entity='Orders',
+        query={"$filter": "ShipCountry eq 'France'",
+               "$orderby": "Freight desc",
+               "$skip": 50,
+               "$top": 3})
+    df = odata_connector.get_df(data_source)
```
adimascio: Maybe we should wrap this in a `try / except socket.error` (or whatever exception fits) in case the service is unreachable and maybe use `pytest.skip` in this case.

toucan_connectors/odata/odata_connector.py
```diff
@@ -0,0 +1,24 @@
...
+    data_source_model: ODataDataSource
+
+    username: str
+    password: str
+    url: str
+
+    def get_df(self, data_source: ODataDataSource) -> pd.DataFrame:
+        Service = ODataService(self.url, reflect_entities=True)
+        Entities = Service.entities[data_source.entity]
+        data = Service.query(Entities).raw(data_source.query)
```
adimascio: A general comment here, not specific to this commit, but maybe we should use a llibrary such as https://github.com/jd/tenacity for network-based queries. Network is by essence unreliable and `tenacity` makes it easy perform multiple tries on failure.

toucan_connectors/__init__.py
piotch: fixed

PR #50, opened 2018-11-11
makefile :: create a new connector from templates

piotch:
```
$ make new_connector type=Magento
```

just makes it easier to start on a new connector

- [X] uses m4 :)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -64,3 +64,10 @@ upload:
 list:
 	@grep '^[^\.#[:space:]].*:' Makefile | \
 		cut -d':' -f1
+
+new_connector:
+	mkdir toucan_connectors/$(name)
+	touch toucan_connectors/$(name)/__init__.py
+	m4 -Dname=$(name) templates/connector.py > toucan_connectors/$(name)/$(name)_connector.py
+	mkdir tests/$(name)
+	m4 -Dname=$(name) templates/tests.py > tests/$(name)/test_$(name).py
--- a/Makefile
+++ b/Makefile
@@ -66,11 +66,13 @@ list:
 		cut -d':' -f1
 
 new_connector:  # $ make new_connector type=Magento
-	$(eval MODULE=`echo "$(type)" | tr A-Z a-z`)
+ ifeq (${type},)
+	@echo -e "Missing type option:\n\tmake $@ type=Magento"
+	exit 1
+ endif
+	${eval MODULE=`echo "${type}" | tr A-Z a-z`}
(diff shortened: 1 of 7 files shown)


piotch:
#https://github.com/ToucanToco/toucan-connectors/issues/46

Makefile
```diff
@@ -64,3 +64,11 @@ upload:
 list:
 	@grep '^[^\.#[:space:]].*:' Makefile | \
 		cut -d':' -f1
+
+new_connector:  # $ make new_connector type=Magento
```
labynocle: to avoid issue if `type` is empty you can add the following Makefile test:

```
ifeq ($(type),)
        @echo -e "Missing type option:\n\tmake $@ type=MyConnecter"
        exit 1
endif
```

otherwise that seems nice :+1: 

nota: we should decide if we use `$[type}` or `$(type)` in our bash variable :D previously William use one approach and I use the other ^^ we never take a decision about this point... :P

Makefile
```diff
@@ -64,3 +64,11 @@ upload:
...
+	$(eval MODULE=`echo "$(type)" | tr A-Z a-z`)
+	mkdir toucan_connectors/${MODULE}
+	touch toucan_connectors/${MODULE}/__init__.py
+	m4 -Dtype=$(type) templates/connector.py.m4 > toucan_connectors/${MODULE}/${MODULE}_connector.py
+	mkdir tests/${MODULE}
+	m4 -Dtype=$(type) templates/tests.py.m4 > tests/${MODULE}/test_${MODULE}.py
```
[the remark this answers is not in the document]
piotch: great, thanks I added the echoes. 

re: doc, we have a script already that we are running once the connector is developed to generate the documentation (cf. https://github.com/ToucanToco/toucan-connectors/blob/master/doc/generate.py)

Makefile
piotch: should I go for `{VAR}` ? is it the one you are using ? 

I tried to copy and paste the parameter checking code above but I must have make a mistake, it gave me : 

```
make new_connector type=Denodo
ifeq (Denodo,)
/bin/bash: -c: line 0: syntax error near unexpected token `Denodo,'
/bin/bash: -c: line 0: `ifeq (Denodo,)'
```

labynocle: I prefer to use the `${VAR}` but I will use what you want :) we just need to use the same

about the `ifeq`, it's not a bash part but a Makefile part ^^ this is why
please refer to the following example: https://github.com/ToucanToco/latapult/blob/master/Makefile#L25

PR #51, opened 2018-11-12
Magento :: add connector

piotch:
Basic support for now, may need to grow up : undelying lib seems fragile and not very usefull, could be tested with a container also.

### DIFF
--- a/tests/magento/test_magento.py
+++ b/tests/magento/test_magento.py
@@ -1,0 +1,29 @@
+# TODO
+# I will test this with a container, probably using this
+# https://hub.docker.com/r/magento/magento2devbox-web/
+
+import pytest
+
+from toucan_connectors.magento.magento_connector import MagentoConnector, MagentoDataSource
+
+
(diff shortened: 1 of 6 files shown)


tests/magento/test_magento.py
```diff
@@ -0,0 +1,27 @@
...
+# https://hub.docker.com/r/magento/magento2devbox-web/
+
+from toucan_connectors.magento.magento_connector import MagentoConnector, MagentoDataSource
+
+
+def test_get_df(mocker):
+
+    mocker.patch('toucan_connectors.magento.magento_connector.API.call')
+
+    data_source = MagentoDataSource(**{'domain': 'test', 'name': 'magento',
```
adimascio: What about using plain named parameters here?

```python

    data_source = MagentoDataSource(domain='test', name='mangento'
                                    resource_path='orders_sale',
                                    arguments=[[{'status': 'pending'}]])

```

PR #78, opened 2019-05-02
new connector :: trello

rhuille:
### Provide a trello connector

Check https://developers.trello.com/docs/get-started for the official API documentation.

The connector allow you to list in a dataset all the cards of a board with a chosen set of "fields".
The available fields are listed in the `Field ` object.

The dataset has one line by card and one columns by fields

exemple of configuration : 
```
DATA_PROVIDERS:[
    name: 'trello',
    key_id: '<your_key_id>',
    token: '<your_token>'
]
DATA_SOURCES:[
        board_id: 'fhUR3kVQ',
        name: 'trello',
        domain: 'my_domain'
]

### CODE REVIEW

File: toucan_connectors/trello/trello_connector.py [hunk 1]
```diff
@@ -0,0 +1,141 @@
...
+
+
+class TrelloConnector(ToucanConnector):
+    type = "Trello"
+    data_source_model: TrelloDataSource
+
+    key_id: str = ''
+    token: str = ''
+
+    def get(self, path, **customParams):
```
piotch: should have a more specific name `get_board` maybe ?

rhuille: Yes good idea !

tests/trello/mock_trello_api_json_responses.py
```diff
@@ -0,0 +1,286 @@
+mock_trello_api_json_responses = {
```
piotch: In this project the fixtures are in JSON files rather than python objects. Cf. https://github.com/ToucanToco/toucan-connectors/tree/master/tests/micro_strategy/fixtures

toucan_connectors/trello/trello_connector.py
```diff
@@ -0,0 +1,148 @@
...
+                                 Fields.lists, Fields.members,  Fields.labels]
+    custom_fields: bool = True
+
+
+class TrelloConnector(ToucanConnector):
+    type = "Trello"
+    data_source_model: TrelloDataSource
+
+    key_id: str = ''
+    token: str = ''
```
piotch: I'm not sure why `key_id` and `token` have default values ? To me they should be required, and have no default ?

rhuille: You can request public board for which `key_id ` and `token` are empty

toucan_connectors/trello/trello_connector.py
```diff
@@ -0,0 +1,148 @@
...
+
+    def replace_id_by_value(self, card_with_id, lists=None, labels=None,
+                            members=None, custom_fields=None):
+        """
+        `card_with_id` is a dictionary containing all data of a card,
+        but with unreadlable id instead of value
+        This fonction return `card_with_value` with the same dictionnary
+        as `card_with_id` but with readable value
+
+        `lists`: dictionnary of correspondance between list names and ids
```
piotch: The name of this parameter is a bit confusing : `lists` sounds like a list of lists, but it is actually a dict, call it `names_ids_mapping` for example and then `members_ids_mapping`

toucan_connectors/trello/trello_connector.py
```diff
@@ -0,0 +1,148 @@
...
+        `members`: dictionnary of correspondance between members names and ids
+        `custom_fields`: dictionnary of correspondance between custom field and there representation
```
piotch:
```suggestion
        `custom_fields`: dictionnary of correspondance between custom field and their representation
```

toucan_connectors/trello/trello_connector.py
```diff
@@ -0,0 +1,148 @@
...
+        """
```
piotch: the `labels` kwarg is not documented, I would either document all or type all and document none.

toucan_connectors/trello/trello_connector.py
```diff
@@ -0,0 +1,148 @@
...
+                    if card_custom_field['value']['checked'] == 'true':
+                        card_with_value[custom_field['name']] = True
+                elif custom_field['type'] == 'list':
+                    list_of_options = {x["id"]: x for x in custom_field['options']}
+                    option = list_of_options[card_custom_field['idValue']]
+                    card_with_value[custom_field['name']] = option['value']['text']
+                else:
+                    warnings.warn(f"""The custom field {custom_field['name']} used into \
+                    your board is of type {custom_field['type']}.
+                    This type is not handled by this app and is simply ignored.""")
```
piotch: I think this code block from 82:100 makes the method too long : 

- at least it should be put in a separate function that deals with custom fields
- could be simplified with a mapping type -> function, I can help with this

toucan_connectors/trello/trello_connector.py
```diff
@@ -0,0 +1,148 @@
...
+                if custom_field['type'] == 'number':
+                    card_with_value[custom_field['name']] = float(
+                        card_custom_field['value']['number'])
+                elif custom_field['type'] == 'text':
+                    card_with_value[custom_field['name']] = card_custom_field['value']['text']
+                elif custom_field['type'] == 'date':
+                    card_with_value[custom_field['name']] = card_custom_field['value']['date']
+                elif custom_field['type'] == 'checkbox':
+                    if card_custom_field['value']['checked'] == 'true':
+                        card_with_value[custom_field['name']] = True
```
piotch: why is the "unchecked" case not covered ?

tests/trello/mock_trello_api_json_responses.py
rhuille: Done !

toucan_connectors/trello/trello_connector.py
rhuille: Let's document all

toucan_connectors/trello/trello_connector.py
```diff
@@ -0,0 +1,148 @@
...
+        return card_with_value
+
+    def get_df(self, data_source: TrelloDataSource) -> pd.DataFrame:
+        # get board caracteristics
+        # the following dictionaries are of the form:
+        # - keys: id of field
+        # - values: readable value of field
+
+        fields_for_request = []
+        lists, labels,  members, custom_fields = None, None, None, None
```
piotch:
```suggestion
        lists, labels, members, custom_fields = None, None, None, None
```

toucan_connectors/trello/trello_connector.py
```diff
@@ -0,0 +1,148 @@
...
+                     for x in self.get_board(f'{data_source.board_id}/lists', fields='name')}
+        if 'labels' in data_source.fields_list:
+            fields_for_request += ['labels']
+            labels = {x['id']: x['name']
+                      for x in self.get_board(f'{data_source.board_id}/labels', fields='name')}
+        if 'members' in data_source.fields_list:
+            fields_for_request += ['idMembers']
+            members = {x['id']: x['fullName']
+                       for x in self.get_board(f'{data_source.board_id}/members', fields='fullName')}
+
```
piotch: the code 113:128 mixes to things, building `fileds_for_request` and setting the value of lists, labels, members, custom fields.

I think we should keep the terms used in the third pary API so if the term is `idList` it should not be called `lists` here in Toucan Connectors, because it adds confusion. In this case you can simply `state that fields_for_request = data_source.fields_list` which is much simpler.

Then the code to set lists, labels, members, custom fields is always the same, you can simplify it with a mapping.

toucan_connectors/trello/trello_connector.py
```diff
@@ -0,0 +1,148 @@
...
+            fields=fields_for_request,
+            customFieldItems='true' if data_source.custom_fields else 'false'
+        )
+
+        # replace all id in `cards_with_id` by the corresponding readable value
+        cards_with_value = [self.replace_id_by_value(card_with_id, lists,
+                                                     labels, members, custom_fields)
+                            for card_with_id in cards_with_id]
+
+        data = pd.DataFrame(cards_with_value)
```
piotch:
```suggestion
        return pd.DataFrame(cards_with_value)
```

toucan_connectors/trello/trello_connector.py
```diff
@@ -0,0 +1,148 @@
...
+
+class TrelloConnector(ToucanConnector):
+    type = "Trello"
+    data_source_model: TrelloDataSource
+
+    key_id: str = ''
+    token: str = ''
+
+    def get_board(self, path, **customParams):
+        return requests.get(f'https://api.trello.com/1/boards/{path}', params={
```
piotch: detail but `https://api.trello.com/1` should be a constant on top of the file

toucan_connectors/trello/trello_connector.py
rhuille: Because when the case is unchecked the response does not return the customField into the list : card_with_id['customFieldItems'].

rhuille: This means that if I want to put `False` you must do another loop on `custom_fields_id_mapping`

toucan_connectors/trello/trello_connector.py
piotch: OK thanks I missed this, could we use `None` as a default then rather than the empty string, a detail but, more redable to me.

rhuille: 👍

toucan_connectors/trello/trello_connector.py
```diff
@@ -81,24 +94,8 @@ def replace_id_by_value(self, card_with_id, lists_ids_mapping=None, labels_id_ma
...
-                elif custom_field['type'] == 'list':
-                    list_of_options = {x["id"]: x for x in custom_field['options']}
-                    option = list_of_options[card_custom_field['idValue']]
-                    card_with_value[custom_field['name']] = option['value']['text']
-                else:
-                    warnings.warn(f"""The custom field {custom_field['name']} used into \
-                    your board is of type {custom_field['type']}.
-                    This type is not handled by this app and is simply ignored.""")
+                card_with_value[custom_field['name']] =\
+                    CUSTOM_FIELD_TYPE_MAPPING[custom_field['type']](card_custom_field, custom_field)
```
piotch: merci 👍 tu gères!!!

toucan_connectors/trello/trello_connector.py
piotch: ok

PR #80, opened 2019-05-23
Implement a retry policy on connectors.

adimascio:
By default, behaviour is unchanged, there's no retry but each connector
now accepts a `retry_policy` paramater that supports the following
customization:

- `max_attempts`: the maximum number of retries before giving up
- `max_delay`: delay, in seconds, above which we should give up
- `wait_time`: time, in seconds, between each retry.

For instance, you can create a new connector that will retry at most 5
times if `get_df` fails by doing:

```python
  my_connector = MyConnector(name=u'foo', retry_policy={'max_attempts': 5})
```

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -28,8 +28,9 @@ extras_require = {
 extras_require['all'] = sorted(set(sum(extras_require.values(), [])))
 
 install_requires = [
+    'pydantic',
+    'tenacity',
     'toucan_data_sdk',
-    'pydantic'
 ]
 
 classifiers = [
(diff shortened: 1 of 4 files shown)


PrettyWood:
[ETQU de connecteur, j'ai une façon plus robuste de retrieve des data (retries)](https://trello.com/c/c1lZqsOg/3285-etqu-de-connecteur-jai-une-fa%C3%A7on-plus-robuste-de-retrieve-des-data-retries)

PR #81, opened 2019-05-28
Feature :: add methods get_df_and_count + explain

virobidou:
Add methods 
* `get_df_and_count` : retrieves a part of the data and the total count

* `explain`: returns metrics about the query

### DIFF
--- a/toucan_connectors/mongo/mongo_connector.py
+++ b/toucan_connectors/mongo/mongo_connector.py
@@ -119,7 +119,7 @@ class MongoConnector(ToucanConnector):
             facet['$facet']['df'] = [{'$limit': limit}]
             data_source.query.append(facet)
             res = self.execute_query(data_source).next()
-            count = res['count'][0]['value'] if len(res['count']) > 1 else 0
+            count = res['count'][0]['value'] if len(res['count']) > 0 else 0
             df = pd.DataFrame(res['df'])
         else:
             df = self.get_df(data_source)
--- a/toucan_connectors/mongo/mongo_connector.py
+++ b/toucan_connectors/mongo/mongo_connector.py
@@ -1,14 +1,16 @@
 import re
+from jq import jq
 from typing import Union
 from urllib.parse import quote_plus
 
 import pandas as pd
 import pymongo
+from bson.son import SON
 from pydantic import validator
 
-from toucan_connectors.toucan_connector import ToucanConnector, ToucanDataSource
 from toucan_connectors.common import nosql_apply_parameters_to_query
-
+from toucan_connectors.toucan_connector import ToucanConnector, \
+    ToucanDataSource
 
 PARAM_PATTERN = r'%\(\w*\)s'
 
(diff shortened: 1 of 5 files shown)


virobidou:
[ETQUser, mes permissions sont appliquées à mon requester report](https://trello.com/c/gQXKMxO9/3246-etquser-mes-permissions-sont-appliqu%C3%A9es-%C3%A0-mon-requester-report)

toucan_connectors/mongo/mongo_connector.py
```diff
@@ -74,29 +91,63 @@ def uri(self):
...
         df = pd.DataFrame(list(data))
-
         client.close()
         return df
 
+    def get_df_and_count(self, data_source, limit):
+        if isinstance(data_source.query, dict):
+            data_source.query = [{'$match': data_source.query}]
+
+        if limit is not None:
```
adimascio: It seems odd to use 2 queries to perform both "get" and "count" since mongo cursors are supposed to keep track of resultset total count even if a `$limit` was applied. Unfortunately, it seems that the `count()` API of the `Cursor` class in the python binding is deprecated in favor of something a bit less useful in our case.

But maybe we could just use `$facet` instead ? For instance:

```python
pipeline = [
    {'$match': {'mycol': {'$gt': 0}}},
    {'$limit': 2},
]

facet = {
    "$facet": {
        "totalCount": [
            {"$count": "count"}
        ],
        "results": pipeline,
    },
}

rset = db['my-collection'].aggregate([facet])
```

will return something like:

```python
[{'results': [{'Label': 'Label 1',
               'mycol': 10,
               '_id': ObjectId('5cdae74e57c59326d22854ba')},
              {'Label': 'Label 2',
               'mycol' 13,
               '_id': ObjectId('5cdae74e57c59326d22854bb')}],
  'totalCount': [{'count': 5}]}]
```

i.e. : there's 5 documents matching the pipeline but since we use `$limit` the resultset itself only contains 2 items.

toucan_connectors/mongo/mongo_connector.py
```diff
@@ -74,29 +91,63 @@ def uri(self):
...
+            else:
+                count = 0
+            data_source.query.append({'$limit': limit})
+
+        df = self.get_df(data_source)
+
+        if limit is None:
+            count = len(df)
+
+        return df, count
```
adimascio: Proposition: use a named tuple or a dict instead?

toucan_connectors/toucan_connector.py
```diff
@@ -36,3 +36,16 @@ def __init_subclass__(cls):
     @abstractmethod
     def get_df(self, data_source: ToucanDataSource) -> pd.DataFrame:
         """Main method to retrieve a pandas dataframe"""
+
+    def get_df_and_count(self, data_source: ToucanDataSource, limit: int):
```
adimascio: The mongo connector implementation suggested that `limit` could be None. Is this actually the case? If so, the type annotation should probably be updated.

PR #82, opened 2019-06-02
Get status and all connectors

PrettyWood:
Add `get_status` and a constant for all existing connectors
(required for connector's interface)

https://github.com/ToucanToco/laputa/pull/2544

### DIFF
--- a/tests/mongo/test_mongo.py
+++ b/tests/mongo/test_mongo.py
@@ -253,17 +253,16 @@ def test_status_bad_host(mongo_connector):
 
 def test_status_bad_port(mongo_connector):
     mongo_connector.port += 1
-    assert mongo_connector.get_status() == {
-        'status': False,
-        'details': [
-            ('Hostname resolved', True),
-            ('Port opened', False),
-            ('Host connection', None),
-            ('Authenticated', None),
-            ('Database available', None)
-        ],
-        'error': '[Errno 61] Connection refused'
-    }
+    status = mongo_connector.get_status()
+    assert status['status'] is False
+    assert status['details'] == [
+        ('Hostname resolved', True),
+        ('Port opened', False),
+        ('Host connection', None),
+        ('Authenticated', None),
+        ('Database available', None)
+    ]
+    assert 'Connection refused' in status['error']
 
 
 def test_status_bad_port2(mongo_connector):
--- a/tests/mongo/test_mongo.py
+++ b/tests/mongo/test_mongo.py
@@ -223,66 +223,105 @@ def test_normalize_query():
 
 
 def test_status_all_good(mongo_connector):
-    assert mongo_connector.get_status() == [
-        ('Hostname resolved', True),
-        ('Port opened', True),
-        ('Host connection', True),
-        ('Authenticated', True),
-        ('Database available', True)
-    ]
+    assert mongo_connector.get_status() == {
+        'status': True,
+        'details': [
+            ('Hostname resolved', True),
+            ('Port opened', True),
+            ('Host connection', True),
+            ('Authenticated', True),
+            ('Database available', True)
+        ],
+        'error': None
+    }
 
 
 def test_status_bad_host(mongo_connector):
     mongo_connector.host = 'localhot'
-    assert mongo_connector.get_status() == [
-        ('Hostname resolved', False),
-        ('Port opened', None),
-        ('Host connection', None),
-        ('Authenticated', None),
-        ('Database available', None)
-    ]
+    assert mongo_connector.get_status() == {
+        'status': False,
+        'details': [
+            ('Hostname resolved', False),
+            ('Port opened', None),
+            ('Host connection', None),
+            ('Authenticated', None),
+            ('Database available', None)
+        ],
+        'error': '[Errno -2] Name or service not known'
+    }
 
 
 def test_status_bad_port(mongo_connector):
     mongo_connector.port += 1
-    assert mongo_connector.get_status() == [
-        ('Hostname resolved', True),
-        ('Port opened', False),
-        ('Host connection', None),
-        ('Authenticated', None),
-        ('Database available', None)
-    ]
+    assert mongo_connector.get_status() == {
+        'status': False,
+        'details': [
+            ('Hostname resolved', True),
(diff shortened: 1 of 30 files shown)


toucan_connectors/__init__.py
```diff
@@ -47,4 +47,28 @@
 with suppress(ImportError):
     from .toucan_toco.trello_connector import TrelloConnector
 
+ALL_CONNECTOR_TYPES = [
```
piotch: could you explain why we have to hardcode this ?

toucan_connectors/toucan_connector.py
```diff
@@ -36,3 +36,7 @@ def __init_subclass__(cls):
     @abstractmethod
     def get_df(self, data_source: ToucanDataSource) -> pd.DataFrame:
         """Main method to retrieve a pandas dataframe"""
+
+    def get_status(self) -> str:
```
piotch: I think we should discuss this, to me me a str is not going to be very useful, I think we should return an array of error objects such as pseudo code:  "[(true, I can speak to localhost:port), (false, I cannot query it)]" or "[(false, I cannot speak to localhost:port), (false, cannot query it)]"

toucan_connectors/__init__.py
```diff
@@ -1,50 +1,55 @@
...
+def _import_connector(con_path, con_class):
+    try:
+        mod = import_module(f'.{con_path}', 'toucan_connectors')
+        con_type = getattr(mod, con_class).type
+    except ImportError:
+        # Retrieve the connector type by parsing the file
+        con_file_path = f'{HERE}/{con_path.replace(".", "/")}.py'
+        with open(con_file_path) as f:
+            con_code = f.read()
+        con_type = re.search(con_class + r'\(ToucanConnector\):.*type = [\'"](\w+)[\'"]', con_code, re.S).group(1)
```
piotch: It's short and sweet with a regex, but I'm worried it's too brittle and covering all the cases would make the regex too complex (here for example we are not catching `type='truc'` (no spaces) or different order of attributes etc...

In this case I don't think that it would be overkill to use ast : 


```
import ast  # std lib
from astsearch import ASTPatternFinder, prepare_pattern  # requires pip install astsearch

connector_pattern = prepare_pattern('class ??(ToucanConnector): ??')
type_pattern = prepare_pattern('type=?')

def first_finder(_ast, pattern):
    return next(ASTPatternFinder(pattern).scan_ast(_ast))

with open('hive.py') as f:
    module = ast.parse(f.read())

connector = first_finder(module, connector_pattern)
print(first_finder(connector, type_pattern).value.s)

>> Hive

```

adimascio: +1 for using `ast` in a way or another (either directly from stdlib, or using [astroid](https://astroid.readthedocs.io/en/latest/) or [astsearch](https://pypi.org/project/astsearch/))

adimascio: Trying to put thoughts in this... we shouldn't have to use such "advanced" techniques for such a simple usage. Why don't we simply register the name (the `type`) of the connector in the `__init__` ? 

I think we were lying in the middle of a bridge between :

- an "automatic plugin" world where connectors are isolated and autonomous pieces of code
  that would "declare" themselves, The 2 big advantages of this approach being that all
  information is clearly encapsulated in the connector and that it could even be registered
  from outside this package;

- a "controlled" catalogue of connectors where a main module / function would statically know
  where to find plugins, try to load and register them.

It looks like we've implemend a strange mix of those 2 approaches but therefore without getting all the benefits. If we want to maintain an explicit catalogue in order to know which connector is available, then let's do that, and simply do something like (pseudo-code):

```python

CONNECTORS_CATALOGUE = {
  'hive': 'toucan_connectors.hive.HiveConnector`,
  'graphql: 'toucan_connectors.graphql.GraphqlConnector',
  # ...
}

AVAILABLE_CONNECTORS = {}

for connector_type, connector_module in CONNECTORS_CATALOGUE.items():
    try:
        connector = __import__(connector_module)
    except ImportError:
        logging.info('conector %s not available', connector_type)
    else:
        AVAILABLE_CONNECTORS[connector_type] = conncetor
```

What do you think?

piotch: yes, sounds good. 

I think we should still add a check that a key of `CONNECTORS_CATALOGUE` is indeed the values of the the attribute `ToucanConnector.type` in the relevant class. This `type` string is going to be written in two places (`CONNECTORS_CATALOGUE` and `ToucanConnector.type`) and it is easy to get confused or make a typo.

adimascio: > it is easy to get confused or make a typo.

Indeed and I would then argue to remove the latter :)

piotch: > Indeed and I would then argue to remove the latter :)

remove `ToucanConnector.type` ? 

I would indeed prefer a solution where there is only one declaration.

tests/mongo/test_mongo.py
```diff
@@ -220,3 +220,69 @@ def test_normalize_query():
 
     query = {'city': 'Test'}
     assert normalize_query(query, {}) == [{'$match': {'city': 'Test'}}]
+
+
+def test_status_all_good(mongo_connector):
+    assert mongo_connector.get_status() == [
```
fspot: ah ok je pige l'idée de la liste :+1: c'est top !

toucan_connectors/toucan_connector.py
```diff
@@ -183,3 +183,31 @@ def get_df_and_count(self, data_source: ToucanDataSource, limit: Union[int, None
...
+        """Check if a hostname is resolved"""
+        try:
+            socket.gethostbyname(hostname)
+            return True
+        except socket.error:
+            return False
+
+    @staticmethod
+    def check_port(host, port) -> bool:
+        with socket.socket(socket.AF_INET, socket.SOCK_STREAM) as s:
```
adimascio: Don't you think we should catch any exception here, despite the usage of `connect_ex` ? AFAIK, you could still have exceptions, for instance if port is not a valid value (e.g. a port out of `unsigned short` range)?

PrettyWood: Done !

PR #86, opened 2019-06-27
microstrat :: improvements

fspot:
## To-Do 
- [x] documentation
- [x] tests

## get full dataset (not just 100 first results)
→ use `limit` option (`-1` == all rows)
```python
ds = MicroStrategyDataSource(
    name="microtest",
    domain="foo",
    dataset="cube",
    id="6137E0964C68D84F107816AA694C2209"
    limit=-1,
)
```

## filter dataset
→ use `viewfilter` option (cf. [usage](https://lw.microstrategy.com/msdz/MSDL/GARelease_Current/_GARelease_Archives/1010/docs/projects/RESTSDK/Content/topics/REST_API/REST_API_ViewFilter.htm))
:sparkles: microstrategy's viewFilter does not natively support filtering by attribute or metric name, but we provide "syntaxic sugar" for them.
**Example** : instead of writing : 
```python
ds = MicroStrategyDataSource(
    name="microtest",
    domain="foo",
    dataset="cube",
    id="6137E0964C68D84F107816AA694C2209",
    viewfilter={
        "operator": "Equals",
        "operands": [{
            "type": "form",
            "attribute": {
                "id": "8D679D3511D3E4981000E787EC6DE8A4"
            },
            "form": {
                "id": "CCFBE2A5EADB4F50941FB879CCF1721C"
            }
        }, {
            "type": "constant",
            "dataType": "Char",
            "value": "Miami"
        }]
    }
)
```
You can write the equivalent : 
```python
ds = MicroStrategyDataSource(
    name="microtest",
    domain="foo",
    dataset="cube",
    id="6137E0964C68D84F107816AA694C2209",
    viewfilter={
        "operator": "Equals",
        "operands": [
            {"attribute": "Call Center@DESC"},
            {"constant": "Miami"}
        ]
    }
)
```

(`constant` / `attribute` / `metric` are expanded, documentation coming soon)

## ability to list available reports & cubes
→ use `dataset="search"` option : 
```python
ds = MicroStrategyDataSource(
    name="microtest",
    domain="foo",
    dataset="search",
    id="revenue analysis",  # id is optional here, and allows to filter by pattern
)
```


toucan_connectors/micro_strategy/data.py
```diff
@@ -1,3 +1,79 @@
+from copy import deepcopy
+from functools import singledispatch
```
piotch: 🎉

PR #96, opened 2019-07-19
Feature :: get form

PrettyWood:
⚠️ If approved, this PR will need a version bump. On laputa, we will need a dedicated PR with the version bump of toucan-connectors and the migration for all the modified connectors

### DIFF
--- a/toucan_connectors/mssql/mssql_connector.py
+++ b/toucan_connectors/mssql/mssql_connector.py
@@ -7,6 +7,9 @@ from toucan_connectors.toucan_connector import ToucanDataSource, ToucanConnector
 
 
 class MSSQLDataSource(ToucanDataSource):
+    # By default SQL Server selects the database which is set
+    # as default for specific user
+    database: str = None
     query: constr(min_length=1)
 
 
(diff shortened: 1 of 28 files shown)


toucan_connectors/mssql/mssql_connector.py
```diff
@@ -18,17 +21,15 @@ class MSSQLConnector(ToucanConnector):
...
     host: str
     user: str
-    db: str = None
     password: str = None
     port: int = None
     connect_timeout: int = None
 
-    @property
-    def connection_params(self):
+    def get_connection_params(self, database):
```
adimascio: General note: shouldn't we define `get_connection_params(database)` (or even `datasource` ?) as the official API and deprecate the `connection_params` property ?

virobidou:
[ETQUser, mes permissions sont appliquées à mon requester report](https://trello.com/c/gQXKMxO9/3246-etquser-mes-permissions-sont-appliqu%C3%A9es-%C3%A0-mon-requester-report)

toucan_connectors/mssql/mssql_connector.py
PrettyWood: Well in most cases you don't need to know the database to connect. But it could be a `get_connection_params(**kwargs)` method indeed

PrettyWood: Done in https://github.com/ToucanToco/toucan-connectors/pull/96/commits/9b90afcb8e2060c32ad1f55b3f34e7e2bde9d8f2

PR #106, opened 2019-09-12
MongoConnector :: keep the client alive and cache validate methods

PrettyWood:


PrettyWood:
@adimascio what do you mean by updating the commit message?

adimascio:
> @adimascio what do you mean by updating the commit message?

You wrote:
```
Now we keep the connector can be used as a context manager
and the validate methods have a small cache
```
I think there are a few missing words :)

PR #111, opened 2019-09-20
Fix :: get_slice with no limit

rhuille:
*Before*
when `limit` is `none`, the addition `offset+limit` raised an error

*now*
when `limit` is `none` it is not apply


toucan_connectors/toucan_connector.py
```diff
@@ -231,7 +231,10 @@ def get_slice(
         Exemple: if offset = 5 and limit = 10 then 10 results are expected from 6th row
         """
         df = self.get_df(data_source, permissions)
-        return DataSlice(df[offset:offset+limit], len(df))
+        if limit:
```
PrettyWood:
```suggestion
        if limit is not None:
```
If I put `limit=0`, I expect an empty dataframe. Not all the results

rhuille: Yes thanks !

PR #115, opened 2019-09-23
Mongo :: get_df_with_regex method

davinov:
Implement a `get_df_with_regex` method, to be able to search in data.

---
Before discussions (https://github.com/ToucanToco/toucan-connectors/pull/115#issuecomment-536872924): 

Implements a `search` method, based on these rules:
https://jeremymikkola.com/posts/2019_03_19_rules_for_autocomplete.html
- [x] Exact matches (e.g.: "ab" => "ab")
- [x] Prefix matches (e.g.: "ab" => "ab.")
- [x] Substring matches (e.g.: "ab" => ".ab.")
- [x] Subsequence start matches (e.g.: "ab" => "a.b.")
- [x] Subsequence middle matches (e.g.: "ab" => ".a.b")
- [x] Alphabetical order
- [x] Shortest match first
- [x] Case insensitive

Search proceeds using one query for each rule. It stops as soon as the number of required results (`limit`) has been reached.

This will be useful for database discovery with autocomplete.

### DIFF
--- a/toucan_connectors/mongo/mongo_connector.py
+++ b/toucan_connectors/mongo/mongo_connector.py
@@ -1,9 +1,10 @@
 from functools import _lru_cache_wrapper, lru_cache
-from typing import Optional, Union
+from typing import Optional, Pattern, Union
 from urllib.parse import quote_plus
 
 import pandas as pd
 import pymongo
+from bson.regex import Regex
 from bson.son import SON
 from cached_property import cached_property
 from jq import jq
@@ -219,6 +220,23 @@ class MongoConnector(ToucanConnector):
             df = self.get_df(data_source, permissions)
             total_count = len(df)
         return DataSlice(df, total_count)
+
+    def get_df_with_regex(
+        self,
+        data_source: MongoDataSource,
+        field: str,
+        regex: Pattern,
+        permissions: Optional[str] = None,
+        limit: Optional[int] = None,
+    ) -> pd.DataFrame:
+        # Create a copy in order to keep the original (deepcopy-like)
+        data_source = MongoDataSource.parse_obj(data_source)
+        data_source.query = normalize_query(data_source.query, data_source.parameters)
+        data_source.query[0]['$match'] = {
+            '$and': [data_source.query[0]['$match']]
+            + [{field: {'$regex': Regex.from_native(regex)}}]
+        }
+        return self.get_slice(data_source, permissions, limit=limit).df
 
     @decorate_func_with_retry
     def explain(self, data_source, permissions=None):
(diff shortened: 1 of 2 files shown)


davinov:
After discussion with @PrettyWood, @piotch and @fspot, we agreed on the fact that this repo must stay as lean as possible, being a thin wrapper to execute queries.
From that, I reworked this PR to only implement a `get_df_with_regex` method in the mongo connector.

The search logic, manipulating a serie of regex requests, is considered application logic. It could vary from one app using thee connectors to another. Therefore, I'll implement this logic in another repo.

toucan_connectors/mongo/mongo_connector.py
```diff
@@ -220,6 +221,25 @@ def get_slice(
             total_count = len(df)
         return DataSlice(df, total_count)
 
+    def get_df_with_regex(
+        self,
+        data_source: MongoDataSource,
+        field: str,
+        regex,
```
PrettyWood:
```suggestion
        regex: str,
```
Obvious but it costs nothing ;) Is there a default to `*` or something?

toucan_connectors/mongo/mongo_connector.py
```diff
@@ -220,6 +221,25 @@ def get_slice(
...
+        # Create a copy in order to keep the original (deepcopy-like)
+        data_source = MongoDataSource.parse_obj(data_source)
+        data_source.query = normalize_query(data_source.query, data_source.parameters)
+        data_source.query[0]['$match'] = {
+            '$and': [data_source.query[0]['$match']]
+            + [{field: {'$regex': Regex.from_native(regex)}}]
+        }
+        if limit is not None:
+            data_source.query.append({'$limit': limit})
+        return self.get_df(data_source, permissions)
```
PrettyWood: It works but I'm afraid we are putting the same logic at different places. I feel like what you did is almost the same as `get_slice` but instead of permissions we add the regex filter.
Wouldn't it be possible to use `get_slice` ?

PrettyWood: As it may be too long to do (and out of scope), I'll probably open an issue if the PR is merged as such.

toucan_connectors/mongo/mongo_connector.py
PrettyWood: (Don't hesitate to push force and rewrite the whole history btw)

toucan_connectors/mongo/mongo_connector.py
davinov: No that's a good idea! I'll do this

PR #120, opened 2019-11-07
Mongo :: instantiate MongoClient with kwargs instead of building a connection string

piotch:
Instantiate `MongoClient` with kwargs rather than with a connection string built from the connector parameters.

It means that it will be easier to add any supported `MongoClient` kwargs at the `MongoConnector` level.

It also means that we can use a connection string in the `host` field and leave out all the other parameters. In the end we can natively support a lot more connections options than we cover with the few parameters that we have explicitly listed in `MongoConnector`. cf. https://docs.mongodb.com/manual/reference/connection-string/

And more specifically this means that when you create a new MongoDB Atlas cluster, you can just copy and paste the connection string that is given by Atlas into our configurations files.

### DIFF
--- a/tests/mongo/test_mongo.py
+++ b/tests/mongo/test_mongo.py
@@ -281,7 +281,6 @@ def test_normalize_query():
 
 
 def test_status_all_good(mongo_connector):
-    # import ipdb; ipdb.set_trace();
     assert mongo_connector.get_status() == {
         'status': True,
         'details': [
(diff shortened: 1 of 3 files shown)


tests/mongo/test_mongo.py
```diff
@@ -305,7 +304,7 @@ def test_status_bad_host(mongo_connector):
         ('Host connection', None),
         ('Authenticated', None),
     ]
-    assert 'not known' in status['error']
+    assert len(status['error']) > 0
```
fspot: Pourquoi avoir changé ce test ?

piotch: le message d'erreur qui est renvoyé dépends de configurations système, et ne contient pas toujours la string "not known"

PR #128, opened 2019-12-23
ROK :: new connector

piotch:
New connector towards ROK Solution (cf. https://www.rok-solution.fr/). The underlying API is GraphQL, but with a "custom" auth mechanism so we make a specific connector for the moment.

### DIFF
--- a/doc/connectors.md
+++ b/doc/connectors.md
@@ -36,6 +36,8 @@
 
 * [PostgresConnector](connectors/postgres.md)
 
+* [ROKConnector](connectors/ROK.md)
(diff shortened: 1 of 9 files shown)


doc/connectors.md
```diff
@@ -36,6 +36,8 @@
 
 * [PostgresConnector](connectors/postgres.md)
 
+* [ROKConnectot](connectors/ROK.md)
```
fspot:
```suggestion
* [ROKConnector](connectors/ROK.md)
```

PR #129, opened 2020-01-14
Bearer integration

PrettyWood:
- The library now handles labels and images for all the connectors (instead of the front-end)
- Added 3 connectors that use Bearer: Google sheets, Aircall and Lightspeed


toucan_connectors/aircall/aircall_connector.py
```diff
@@ -0,0 +1,34 @@
...
+    This is a connector for [Aircall](https://developer.aircall.io/api-references/#endpoints)
+    using [Bearer.sh](https://app.bearer.sh/)
+    """
+
+    data_source_model: AircallDataSource
+    bearer_integration = 'aircall_oauth'
+    bearer_auth_id: str
+
+    def _retrieve_data(self, data_source: AircallDataSource) -> pd.DataFrame:
+        # https://developers.google.com/sheets/api/samples/reading
```
piotch: bad link non ?

PR #131, opened 2020-01-21
Fix/pydantic dependency

QuentinMadura:
**Summary:** 
Update of Pydantic version from 0.31.1 to 1.3.

_Before_
Using Schema when we need to define more stuff on a field than just the type and default.
Using UrlStr to validate url field.

_After_
[Breaking change]: [Update Schema with Field](https://github.com/samuelcolvin/pydantic/issues/577) => Add compatibility schema and schema_model so it can be used easily on a [dataclass](https://docs.python.org/3/library/dataclasses.html).
Update Mongo connector to prevent client parameter generated by Pydantic.
Fix retry_decorator on the toucan connector to prevent duplicate retry_policy.
[Breaking Change]: Remove UrlStr and [rewrite of URL parsing logic](https://github.com/samuelcolvin/pydantic/pull/755).
Lint files.


PrettyWood:
Done in https://github.com/ToucanToco/toucan-connectors/pull/135

PR #140, opened 2020-02-13
Mapping multiple connectors

QuentinMadura:

PR #141, opened 2020-02-19
Fix :: Circle CI

PrettyWood:
upgrade circleci machine because installation of msodbcsql17 fails on ubuntu 14.04


QuentinMadura:
Fixed here: https://github.com/ToucanToco/toucan-connectors/pull/140

PR #150, opened 2020-03-25
Permissions :: change the input format

erenard:
## Checklist


tests/mongo/test_mongo_translator.py
```diff
@@ -1,23 +1,119 @@
...
+            },
+        ]
+    }
+    assert permission_conditions_to_mongo_query(c) == {
+        '$and': [
+            {'country': {'$eq': 'France'}},
+            {'$or': [{'city name': {'$in': ['Paris', 'London']}}, {'population': {'$eq': 42}}]},
+        ]
+    }
+    # looking for 100% code coverage
```
davinov:
```suggestion
    # Invalid and/or condition list
```

tests/mongo/test_mongo_translator.py
```diff
@@ -27,30 +123,43 @@ def test_mongo_expression():
...
-    assert MongoExpression().parse(expr) == expected
-
-    expr = "(type == 'YTD') or (periode == {{periode}})"
-    expected = {'$or': [{'type': 'YTD'}, {'periode': '{{periode}}'}]}
-    assert MongoExpression().parse(expr) == expected
-
-    expr = "(type == 'YTD') or (periode == 'yo_{{my_indic[\"a\"]}}')"
-    expected = {'$or': [{'type': 'YTD'}, {'periode': 'yo_{{my_indic["a"]}}'}]}
-    assert MongoExpression().parse(expr) == expected
+def test_permission_conditions_to_mongo_query_with_jinja():
```
davinov: I have trouble understanding the point of these tests, as we are not interpolating the variables in curly braces here.
@PrettyWood do you know why it was necessary to test that in the first place ? Is that to verify that the values are indeed not interpolated ?

tests/test_pandas_translator.py
```diff
@@ -0,0 +1,71 @@
...
+
+    assert PandasOperatorMapping.EQUAL.to_clause('type', 'YTD') == 'type == YTD'
+    assert PandasOperatorMapping.NOT_EQUAL.to_clause('type', 'YTD') == 'type != YTD'
+    assert PandasOperatorMapping.LOWER_THAN.to_clause('type', 'YTD') == 'type < YTD'
+    assert PandasOperatorMapping.LOWER_THAN_EQUAL.to_clause('type', 'YTD') == 'type <= YTD'
+    assert PandasOperatorMapping.GREATER_THAN.to_clause('type', 'YTD') == 'type > YTD'
+    assert PandasOperatorMapping.GREATER_THAN_EQUAL.to_clause('type', 'YTD') == 'type >= YTD'
+    assert PandasOperatorMapping.IN.to_clause('type', 'YTD') == 'type in YTD'
+    assert PandasOperatorMapping.NOT_IN.to_clause('type', 'YTD') == 'type not in YTD'
+
```
davinov: Maybe we can add `with raises` tests here for unsupported operators?

toucan_connectors/common.py
```diff
@@ -122,7 +122,16 @@ def _handle_missing_params(elt, params):
...
 
-def render_raw_permissions(query, parameters):
+def apply_query_parameters(query, parameters):
+    """
+    Apply parameters to query
+
+    :param      query:       The query
+    :type       query:       { type_description }
+    :param      parameters:  The parameters
+    :type       parameters:  { type_description }
```
davinov: I suggest explaining what the function does here instead of documenting the arguments.
I suggest:
```
Interpolate the query, which is a Jinja templates, with the provided parameters.
```

What we also can do to improve this doc is add type-hints to the signature:
 `def apply_query_parameters(query: str, parameters: dict): -> str`

toucan_connectors/mongo/mongo_translator.py
```diff
@@ -1,105 +1,77 @@
...
 
-from toucan_connectors.common import Column, Expression, Operator, Value
 
+def permission_condition_to_mongo_clause(condition: dict) -> dict:
+    if 'operator' not in condition:
+        raise KeyError('key "operator" is missing from permission condition')
+    else:
+        operator = MongoOperatorMapping.from_identifier(condition['operator'])
+        if operator is None:
+            raise ValueError(f'Unsupported operator:{condition["operator"]}')
```
davinov:
```suggestion
            raise ValueError(f'Unsupported operator: {condition["operator"]}')
```

toucan_connectors/common.py
```diff
@@ -1,13 +1,97 @@
...
+    The `translate` method take the condition and returnthe corresponding part of a query.
+
+    Each available operator supported must be implemented in child classes.
+    """
+
+    def translate(cls, condition):
+        raise NotImplementedError
+
+    @classmethod
+    def EQUAL(cls):
```
PrettyWood: I don't understand why not just do something like this
```python
class ConditionTranslator(ABC):
    @classmethod
    @abstractmethod
    def translate(cls, condition):
         "main method..."

    @classmethod
    @abstractmethod
    def eq(cls, column, value) -> Any:  # return type depends on translators(string for pandas, dict for mongo, ...)
        ...

    def ...
 ```

I think it's way clearer and explicit. What do you think?

PR #162, opened 2020-05-18
[mongo] more verbose explain infos

adimascio:
`aggregate(..., explain=True)` only return basic `queryPlanner`
information (cf. https://docs.mongodb.com/manual/reference/explain-results/<Paste>)

This changes triggers an `explain` command rather than an `aggregate` one
and choose the `executionStats` output mode. This will provide, for each stage,
information such as number of returned elements, winning plans, actual
elapsed time, etc.


testinnplayin:
I approved this since the fails were due to linting issues (yay!!)

PR #170, opened 2020-06-09
Trello connector implements "filter" param

rhuille:
of trello api

### DIFF
--- a/toucan_connectors/trello/trello_connector.py
+++ b/toucan_connectors/trello/trello_connector.py
@@ -44,10 +44,25 @@ class Fields(str, Enum):
     labels = 'labels'
 
 
+class CardsFilter(str, Enum):
+    """
+    A filter to match cards in a specific state on a board.
+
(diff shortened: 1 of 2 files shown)


toucan_connectors/trello/trello_connector.py
```diff
@@ -44,10 +44,23 @@ class Fields(str, Enum):
     labels = 'labels'
 
 
+class Filter(str, Enum):
+    """
+    cf https://developer.atlassian.com/cloud/trello/rest/#api-boards-id-cards-filter-get
+    """
```
davinov, line 50: IMO, this is a very generic class name. May we rename it `CardsFilter`, `CardStatus`, or something like it?

```suggestion
class CardsFilter(str, Enum):
    """
    A filter to match cards in a specific state on a board.
    
    https://developer.atlassian.com/cloud/trello/rest/#api-boards-id-cards-filter-get
    """
```

PR #179, opened 2020-06-30
fix(error message) :: ETL validation crash in HTTP connector

testinnplayin:
## Change Summary

Proposes error message for bad variable in endpoint while configuring datasets retrieved from an HTTP connector

This PR makes the error message more explicit.

## Screen shot

Capture d’écran de 2020-06-30 11-06-54

NOTE: wording has been changed a little in the error message proposed in this PR.

## Checklist

### DIFF
--- a/tests/test_common.py
+++ b/tests/test_common.py
@@ -1,7 +1,7 @@
 import pytest
 
 from toucan_connectors.common import (
-    NonValidEndpointVariable,
+    NonValidVariable,
     apply_query_parameters,
     nosql_apply_parameters_to_query,
 )
@@ -187,6 +187,7 @@ def test_bad_variable_in_query():
     """It should thrown a NonValidEndpointVariable exception if bad variable in endpoint"""
     query = {'url': '/stuff/%(thing)s/foo'}
     params = {}
-    with pytest.raises(NonValidEndpointVariable) as err:
-        nosql_apply_parameters_to_query(query, params)
-    assert str(err.value) == 'Non valid variable thing in endpoint'
+    nosql_apply_parameters_to_query(query, params)
+    with pytest.raises(NonValidVariable) as err:
+        nosql_apply_parameters_to_query(query, params, handle_errors=True)
+    assert str(err.value) == 'Non valid variable thing'
--- a/tests/test_common.py
+++ b/tests/test_common.py
@@ -1,4 +1,10 @@
-from toucan_connectors.common import apply_query_parameters, nosql_apply_parameters_to_query
+import pytest
+
+from toucan_connectors.common import (
+    NonValidEndpointVariable,
+    apply_query_parameters,
+    nosql_apply_parameters_to_query,
+)
 
 
 def test_apply_parameter_to_query_do_nothing():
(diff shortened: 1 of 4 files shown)


testinnplayin:
[AS an app builder, when I interpolate a variable in my url using the HTTP connector I would like the ETL validator not to crash (and I would like to have an understandable error message)](https://trello.com/c/pMmQkrjv/4191-as-an-app-builder-when-i-interpolate-a-variable-in-my-url-using-the-http-connector-i-would-like-the-etl-validator-not-to-crash-a)

toucan_connectors/common.py
```diff
@@ -23,12 +23,17 @@
...
+    """ Error thrown for a non valid variable in endpoint """
+
+
 def nosql_apply_parameters_to_query(query, parameters):
     """
     WARNING : DO NOT USE THIS WITH VARIANTS OF SQL
     Instead use your client library parameter substitution method.
     https://www.owasp.org/index.php/Query_Parameterization_Cheat_Sheet
     """
+    print('first query ', query)
```
davinov, line 36: Have you forgotten to remove this `print`? :)

tests/test_common.py
```diff
@@ -175,3 +181,12 @@ def test_render_raw_permission():
         '(indic0 == "0" or indic1 == 1) and ' 'indic2 == "yo_2" and indic_list == [\'0\', 1, \'2\']'
     )
     assert apply_query_parameters(query, params) == expected
+
+
+def test_bad_variable_in_query():
+    """It should thrown a NonValidEndpointVariable exception if bad variable in endpoint"""
+    query = {'url': '/stuff/%(thing)s/foo'}
+    params = {}
+    with pytest.raises(NonValidEndpointVariable) as err:
```
davinov, line 190: I recently discovered that pytest.raises can match on the content on the expression:
https://docs.pytest.org/en/stable/assert.html#assertions-about-expected-exceptions

It's a nice and concise way to do the same thing, so it's not mandatory to merge this PR

toucan_connectors/common.py
```diff
@@ -103,6 +108,10 @@ def _handle_missing_params(elt, params):
                             Template('{{ %s }}' % m, undefined=StrictUndefined).render(params)
                         except Exception:
                             missing_params.append(m)
+                            if k == 'url':
+                                raise NonValidEndpointVariable(
+                                    f'Non valid variable {m} in endpoint'
+                                )
```
davinov, line 114: This method `_handle_missing_params` is very generic. It's not directly linked to the HTTP connector in which the 'url' param is mandatory.
I would rather add an argument `mandatory_keys` to `_handle_missing_params`. The mandatory keys would not be allowed to fail the template rendering step, and would raise an error instead of ignoring it.
Then, the HTTP connector could use this method, saying that the `url` parameter is not allowed to fail.
What do you think?

toucan_connectors/common.py
testinnplayin: Oops! :blush:

toucan_connectors/common.py
testinnplayin: 'url' is a little bit poorly named, it's actually a partial url like `/posts` rather than `https://example.com/posts`. It's not required.

It's not technically a mandatory key but I can pass an array of parameters with a different name perhaps :smile:

PR #182, opened 2020-07-09
Adding support for DocumentDB with degraded mode #181

riderx:
## Change Summary

This component is a copy of Mongo component where i removed the $facet aggregation.
I tried to remplace it by $lookup but it was too slow, and tried to use $project with $push $$ROOT who of course was the best alternative but not supported in DocumentDB too.

I tryed in last to do two separate request but wasn't optimal too.

So i completly removed the total count in get_slice function 
it's always set as MAX_COUNTED_ROWS
Since it was already capped at this number.

## Related issue number

#181 

## Checklist


piotch:
This creates a lot of duplicated code n'est-ce pas ? Could you not import all that you need from the mongo connector and rewrite only the bits which are different ?

Testing : if you do what I suggested above the only thing you have to test is the re-written code. Which you can test against mongo (juste like we do in the mongo connector tests).

riderx:
Good idea ! i will work on it !

BobCashStory:
@piotch ok for you now ?

toucan_connectors/mongo/mongo_connector.py
```diff
@@ -206,6 +206,28 @@ def get_df(self, data_source, permissions=None):
         data_source.query = apply_permissions(data_source.query, permissions)
         return self._retrieve_data(data_source)
 
+    def is_non_geniune(self):
```
fspot, line 209: s/geniune/genuine/g :)

riderx: Done

fspot:
Closing for now because this adds an overhead of 2 requests at each call to `get_slice` method, which means at each data request on toucan's api using mongo connector (as this connector is used both by external and internal mongo).
If the need for this option raises again, we could use a flag on `MongoConnector` to activate this check.

E.g. : 

```python
check_non_genuine: Optional[bool] = Field(
    None,
    description='Checks wether the mongo server is genuine or not (e.g DocumentDB, CosmosDB)'
)
```

PR #188, opened 2020-08-04
fix(mongo): fix status error format

### LINKED ISSUE #187: Configuration Problem with MongoDB Connector Test in Github Actions
The function `test_status_bad_username` never passes even though this code has not been touched for a while.

What? Seems to be a configuration issue because this test breaks on Github Actions and not locally (at least for @testinnplayin). It breaks for all three Python environments.
What behavior do we expect? That the test passes in all three Python environments.
What version? Since 0.38.0 of toucan-connectors.
Recurrent? Yes.

Capture d’écran de 2020-08-04 09-32-19

davinov:
## Change Summary
Mongo error format seems to differ in some local setups and in the CI (https://github.com/ToucanToco/toucan-connectors/pull/186/checks?check_run_id=941369017)

## Related issue number
Fixes #187 

## Checklist

* ~[ ] Documentation reflects the changes where applicable~

### DIFF
--- a/tests/mongo/test_mongo.py
+++ b/tests/mongo/test_mongo.py
@@ -417,16 +417,15 @@ def test_status_unreachable(mongo_connector, mocker):
 
 def test_status_bad_username(mongo_connector):
     mongo_connector.username = 'bibou'
-    assert mongo_connector.get_status() == {
-        'status': False,
-        'details': [
-            ('Hostname resolved', True),
-            ('Port opened', True),
-            ('Host connection', True),
-            ('Authenticated', False),
-        ],
-        'error': 'Authentication failed.',
-    }
+    status = mongo_connector.get_status()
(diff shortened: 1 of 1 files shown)


PrettyWood:
@davinov It's the same in local and in CI.
It's just the new release of pymongo https://github.com/mongodb/mongo-python-driver/releases/tag/3.11.0
Btw I got a new error in local with `yarl`. So CI will probably break with another issue.

davinov:
@PrettyWood yes it's caused by the pymongo update. The difference between some local setups and the CI is due to the requirements not being exactly pinned (`pymongo>=3.6.1`). So reinstalling in a local env, existing before the release of pymongo 3.11.0, will not necessarily be updated yet still be valid.
That's why I altered a bit the test so it's still valid for any of these versions of pymongo

PR #199, opened 2020-08-28
chore(deps): bump black and re-lint

testinnplayin:
## Change Summary

Bumping version of black and lint the files that need to be linted.

Contains bump of connectors version so can rebuild.


setup.py
```diff
@@ -66,7 +66,7 @@ def get_static_file_paths():
 
 setup(
     name='toucan_connectors',
-    version='0.39.4',
+    version='0.39.5',
```
erenard, line 69: C'est la version de toucan-connecteur, on ne devrai pas y toucher, si ?

erenard (changes requested):

erenard (approved):

PR #201, opened 2020-08-31
fix(gbq): use standard SQL as default query syntax

PrettyWood:
We used to have legacy SQL syntax by default for google big query, which was confusing as the official documentation and most of conceptors are used to writing standard SQL.
This PR changes this by setting standard SQL as the default dialect. Conceptors can still manually select legacy syntax if needed


PrettyWood:
[AS a conceptor, i want to use the standard sql syntax on a google big query connector](https://trello.com/c/xkoXGllt/4293-as-a-conceptor-i-want-to-use-the-standard-sql-syntax-on-a-google-big-query-connector)

PR #207, opened 2020-09-11
chore: added auth_flow_id to Google Sheets 2

testinnplayin:
## Change Summary

This adds a new property to the Google Sheets 2 connector : `auth_flow_id`, which behaves a lot like the old `bearer_auth_id`. It's used as an identifier for the Google Sheets 2 connector instead of the connector's name, which can be modified by the user.

This renders the oauth dance more robust because 1) we no longer rely on a property that can be changed at any moment by the user and 2) it prevents fan out throw several collections when a user updates their connector's name.

### DIFF
--- a/doc/connectors/google_sheets_2.md
+++ b/doc/connectors/google_sheets_2.md
@@ -15,7 +15,7 @@ The `baseroute` is fixed and is 'https://sheets.googleapis.com/v4/spreadsheets/'
 
 The `secrets` dictionary contains the `access_token` and a `refresh_token` (if there is one). Though `secrets` is optional during the initial creation of the connector, it is necessary for when the user wants to make requests to the connector. If there is no `access_token`, an Exception is thrown.
 
-The `auth_flow_id` property is like a identifier that is used to identify the connector.
+The `auth_flow_id` property is like an identifier that is used to identify the secrets associated with the connector.
 
 
 ```coffee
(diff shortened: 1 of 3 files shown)


testinnplayin:
[(3) ETQC, je peux changer le nom du connecteur OAuth que je suis en train de créer](https://trello.com/c/YcjFa3cn/811-3-etqc-je-peux-changer-le-nom-du-connecteur-oauth-que-je-suis-en-train-de-cr%C3%A9er)

doc/connectors/google_sheets_2.md
```diff
@@ -14,6 +15,8 @@ The `baseroute` is fixed and is 'https://sheets.googleapis.com/v4/spreadsheets/'
 
 The `secrets` dictionary contains the `access_token` and a `refresh_token` (if there is one). Though `secrets` is optional during the initial creation of the connector, it is necessary for when the user wants to make requests to the connector. If there is no `access_token`, an Exception is thrown.
 
+The `auth_flow_id` property is like a identifier that is used to identify the connector.
```
davinov, line 18:
```suggestion
The `auth_flow_id` property is like a identifier that is used to identify the secrets associated to the connector.
```

davinov (approved):

doc/connectors/google_sheets_2.md
testinnplayin: Yes j'ai aussi fait une bêtise ... ça devrait être 'is like an identifier'... ne dis pas aux autres :blush:

Issue #206, opened 2020-09-06
Split connector's config model from connector's logic


davinov:
With the arrival of new connectors, like the new google sheets one (#202), I see new uses of connector's properties:
- some "flagging" attributes: like `auth_flow: "oAuth2"`. These attributes are useful to indicate behavior changes for some groups of connectors. 
- some "internal" attributes: like `secrets`. These attributes are necessary to run a query, but should not be filled directly by a user when configuring a connector. Instead, they are populated by the server that makes use of the connector.
Both have in common that they should not appear in the schema (which will be used to display a form) to create or edit a connector.

For the first type, I tested the `ClassVar` annotation and making the attribute private (prepended with `_`). It does prevent to display the info in the schema, but also hide the info when exporting the document as `dict`. Therefore, a front-end would not be able to have this information, and for example open a consent screen for connectors having the authFlow property to "oAuth2".
This starts to show that we want our users to configure, and what we want to export as a connector could sometimes becomes a bit different.

For the second type, private attributes works well combined with setters. But it's still a bit of tricky logic to have to think of making them private to exclude them from being stored.

What I suggest is to separate the `ConnectorConfig` from the `Connector` itself.
`ConnectorConfig` is a good application of pydantic's `BaseModel`. It's something that will comes from user input and stored in a DB or a config file. It should be validated.
`Connector` stores a lot of logic, and this logic needs the configuration object, but also more and more other info.

Users provides the `ConnectorConfig`, which is stored in DB.
Server creates a `Connector` from a `ConnectorConfig` and other relevant info.
Serve can serialize a `Connector` and send it, with its config and other relevant info if necessary.

What are your opinion on this?

Related links:
https://github.com/samuelcolvin/pydantic/issues/655 specially [this comment](https://github.com/samuelcolvin/pydantic/issues/655#issuecomment-610892084) about God-like objects

testinnplayin:
I think it's a good idea to separate out the configuration but not sure in a way about the database from the point of view that so far no connector sends queries to a database/our database. Currently, it's only our server (private repo) that handles communication with a database. We lose the abstraction that the connectors project so far has from any of our other projects.

I'm for keeping them in configuration files and in my mind that's what we should ultimately do (for eg. in JSON files).

PR #208, opened 2020-09-16
feat(gsheets2): added hidden properties

testinnplayin:
## Change Summary

We added a class to hide certain properties from the user. These can be divided into two main categories: flags and non-flags.

Flags are properties that are required by the front to do something. In this case `auth_flow_id` is for initiating the oauth dance.

Non-flags are everything else. Non-flags can be divided into two categories: static and dynamic. Static non-flags are kept on the connector itself and have a '_' at the front of their name. Dynamic ones are kept on the new class. The reason why we put the static non-flags is so that we don't have to return them if we don't want to when we fetch hidden properties from the class that contains them.

On the front-end, we will eventually want to hide all properties that are stored in 'hidden_properties' on the connector. Those properties that begin with a '_' are hidden automatically.

Dynamic properties cannot be hidden with a '_' in front of their variable name because we can't set them afterwards.

## Related issue number

[https://github.com/ToucanToco/toucan-connectors/issues/206](https://github.com/ToucanToco/toucan-connectors/issues/206)

## Checklist

### DIFF
--- a/doc/connectors/google_sheets_2.md
+++ b/doc/connectors/google_sheets_2.md
@@ -4,16 +4,16 @@
 
 * `type`: `"GoogleSheets2"`
 * `name`: str, required
-* `auth_flow`: str
+* `_auth_flow`: str
 * `auth_flow_id`: str
-* `baseroute`: str
-* `secrets`: dict
+* `_baseroute`: str
+* `hidden_properties`: GoogleSheets2HiddenProperties
 
-The `auth_flow` property marks this as being a connector that uses the connector_oauth_manager for the oauth dance.
+The `_auth_flow` property marks this as being a connector that uses the connector_oauth_manager for the oauth dance. It is hidden from being rendered in the front but appears in the front as a flag for initiating the oauth dance. It is a static *flag*.
 
-The `baseroute` is fixed and is 'https://sheets.googleapis.com/v4/spreadsheets/'.
+The `_baseroute` is fixed and is 'https://sheets.googleapis.com/v4/spreadsheets/'. It too is hidden from being rendered in the front. It is a static *non-flag*
 
-The `secrets` dictionary contains the `access_token` and a `refresh_token` (if there is one). Though `secrets` is optional during the initial creation of the connector, it is necessary for when the user wants to make requests to the connector. If there is no `access_token`, an Exception is thrown.
+The `hidden_properties` propertie contains the `access_token` and a `refresh_token` (if there is one) in a class called `GoogleSheets2HiddenProperties`. Though `hidden_properties` is optional during the initial creation of the connector, it is necessary for when the user wants to make requests to the connector. If there is no `access_token`, an Exception is thrown. This is a dynamic *non-flag* property.
 
 The `auth_flow_id` property is like an identifier that is used to identify the secrets associated with the connector.
 
@@ -22,6 +22,7 @@ The `auth_flow_id` property is like an identifier that is used to identify the s
 DATA_PROVIDERS: [
   type:    'GoogleSheets'
   name:    '<name>'
+  auth_flow_id: '<auth_flow_id>'
 ,
   ...
 ]
--- a/toucan_connectors/google_sheets_2/google_sheets_2_connector.py
+++ b/toucan_connectors/google_sheets_2/google_sheets_2_connector.py
@@ -7,7 +7,7 @@ from typing import Any, Dict, Optional
 
 import pandas as pd
 from aiohttp import ClientSession
-from pydantic import Field, create_model
+from pydantic import BaseModel, Field, create_model
 
 from toucan_connectors.common import ConnectorStatus, HttpError, fetch, get_loop
 from toucan_connectors.toucan_connector import ToucanConnector, ToucanDataSource, strlist_to_enum
@@ -43,8 +43,11 @@ class GoogleSheets2DataSource(ToucanDataSource):
         constraints = {}
         with suppress(Exception):
             partial_endpoint = current_config['spreadsheet_id']
-            final_url = f'{connector.baseroute}{partial_endpoint}'
-            data = connector._run_fetch(final_url, connector.secrets['access_token'])
+            final_url = f'{connector._baseroute}{partial_endpoint}'
+            data = connector._run_fetch(
+                final_url,
+                connector.hidden_properties.get_hidden_properties()['secrets']['access_token'],
+            )
             available_sheets = [str(x['properties']['title']) for x in data['sheets']]
             constraints['sheet'] = strlist_to_enum('sheet', available_sheets)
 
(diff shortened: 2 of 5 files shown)


testinnplayin:
[(8) EPIC ETQC, je ne vois pas des propriétés privées dans le formulaire d'édition du connecteur GoogleSheets2](https://trello.com/c/woTQo3Fh/850-8-epic-etqc-je-ne-vois-pas-des-propri%C3%A9t%C3%A9s-priv%C3%A9es-dans-le-formulaire-d%C3%A9dition-du-connecteur-googlesheets2)

doc/connectors/google_sheets_2.md
```diff
@@ -4,16 +4,16 @@
...
 
-The `baseroute` is fixed and is 'https://sheets.googleapis.com/v4/spreadsheets/'.
+The `_baseroute` is fixed and is 'https://sheets.googleapis.com/v4/spreadsheets/'. It too is hidden from being rendered in the front. It is a static *non-flag*
 
-The `secrets` dictionary contains the `access_token` and a `refresh_token` (if there is one). Though `secrets` is optional during the initial creation of the connector, it is necessary for when the user wants to make requests to the connector. If there is no `access_token`, an Exception is thrown.
+The `hidden_properties` propertie contains the `access_token` and a `refresh_token` (if there is one) in a class called `GoogleSheets2HiddenProperties`. Though `hidden_properties` is optional during the initial creation of the connector, it is necessary for when the user wants to make requests to the connector. If there is no `access_token`, an Exception is thrown. This is a dynamic *non-flag* property.
```
davinov, line 16:
```suggestion
The `hidden_properties` property contains the `access_token` and a `refresh_token` (if there is one) in a class called `GoogleSheets2HiddenProperties`. Though `hidden_properties` is optional during the initial creation of the connector, it is necessary for when the user wants to make requests using the connector. If there is no `access_token`, an Exception is thrown. This is a dynamic property (not a static one).
```

doc/connectors/google_sheets_2.md
```diff
@@ -4,16 +4,16 @@
...
-* `baseroute`: str
-* `secrets`: dict
+* `_baseroute`: str
+* `hidden_properties`: GoogleSheets2HiddenProperties
 
-The `auth_flow` property marks this as being a connector that uses the connector_oauth_manager for the oauth dance.
+The `_auth_flow` property marks this as being a connector that uses the connector_oauth_manager for the oauth dance. It is hidden from being rendered in the front but appears in the front as a flag for initiating the oauth dance. It is a static *flag*.
 
-The `baseroute` is fixed and is 'https://sheets.googleapis.com/v4/spreadsheets/'.
+The `_baseroute` is fixed and is 'https://sheets.googleapis.com/v4/spreadsheets/'. It too is hidden from being rendered in the front. It is a static *non-flag*
```
davinov, line 14:
```suggestion
The `_baseroute` is fixed and is 'https://sheets.googleapis.com/v4/spreadsheets/'. It too is hidden from being rendered in the front. It is a static property.
```
Or class property (something that doesn't change with the instance)

doc/connectors/google_sheets_2.md
```diff
@@ -4,16 +4,16 @@
...
-* `auth_flow`: str
+* `_auth_flow`: str
 * `auth_flow_id`: str
-* `baseroute`: str
-* `secrets`: dict
+* `_baseroute`: str
+* `hidden_properties`: GoogleSheets2HiddenProperties
 
-The `auth_flow` property marks this as being a connector that uses the connector_oauth_manager for the oauth dance.
+The `_auth_flow` property marks this as being a connector that uses the connector_oauth_manager for the oauth dance. It is hidden from being rendered in the front but appears in the front as a flag for initiating the oauth dance. It is a static *flag*.
```
davinov, line 12:
```suggestion
The `_auth_flow` property marks this as being a connector that uses the connector_oauth_manager for the oauth dance. It is hidden from being rendered in the front but appears in the front as a flag for initiating the oauth dance. It is a static *property*.
```
I prefer we use the term _property_, which has a clear meaning in Python.
Flag is less clear : it suggest a boolean, a feature-flag, and many other things.

toucan_connectors/google_sheets_2/google_sheets_2_connector.py
```diff
@@ -54,20 +57,41 @@ def get_form(cls, connector: 'GoogleSheets2Connector', current_config):
 Secrets = Dict[str, Any]
 
 
+class GoogleSheets2HiddenProperties(BaseModel):
+    """
+    This class contains dynamic hidden properties for Google Sheets 2 that are not flags
```
davinov, line 62:
```suggestion
    This class contains dynamic hidden properties for Google Sheets 2 that are not configured directly by a user, like tokens retrieved via OAuth2
```

toucan_connectors/google_sheets_2/google_sheets_2_connector.py
```diff
@@ -54,20 +57,41 @@ def get_form(cls, connector: 'GoogleSheets2Connector', current_config):
...
 
+    # dynamic, non-hidden variables
     data_source_model: GoogleSheets2DataSource
-
-    auth_flow = 'oauth2'
-
     auth_flow_id: Optional[str]
 
-    # The following should be hidden properties
+    # flagging variable; required for initiating the oauth dance
```
davinov, line 87: :+1:

toucan_connectors/google_sheets_2/google_sheets_2_connector.py
```diff
@@ -54,20 +57,41 @@ def get_form(cls, connector: 'GoogleSheets2Connector', current_config):
...
+    _auth_flow = 'oauth2'
 
-    baseroute = 'https://sheets.googleapis.com/v4/spreadsheets/'
+    # static variable
+    _baseroute = 'https://sheets.googleapis.com/v4/spreadsheets/'
```
davinov, line 91: Could we make that a real class property by the way ? That will save some db space and prevent to send it to the front-end

toucan_connectors/google_sheets_2/google_sheets_2_connector.py
```diff
@@ -54,20 +57,41 @@ def get_form(cls, connector: 'GoogleSheets2Connector', current_config):
 Secrets = Dict[str, Any]
 
 
+class GoogleSheets2HiddenProperties(BaseModel):
```
davinov, line 60: I don't really see the value of using a Pydantic BaseModel here. Their usage is to validate/enforce a model, when some data comes from an uncertain source, like a form filled by someone, or an unknown database without a schema.
I wonder if a very simple dataclass would be enough here.

toucan_connectors/google_sheets_2/google_sheets_2_connector.py
```diff
@@ -54,20 +57,41 @@ def get_form(cls, connector: 'GoogleSheets2Connector', current_config):
...
+    # dynamic, non-hidden variables
     data_source_model: GoogleSheets2DataSource
-
-    auth_flow = 'oauth2'
-
     auth_flow_id: Optional[str]
 
-    # The following should be hidden properties
+    # flagging variable; required for initiating the oauth dance
+    _auth_flow = 'oauth2'
```
davinov, line 88: I took a few minutes to look at Pydantic's docs and saw the keep_untouched config option.
https://pydantic-docs.helpmanual.io/usage/model_config/

Is that worth to try ?

davinov (changes requested):
Thanks for all this work and research you put it in.

However, reading the code, I feel weird  about the `hidden_properties` method.
First : it contains only a token for now, so maybe a generic name is not appropriate to make following readers understand what this is about. IMO this name would have make sense if this was in the base class `ToucanConnectors` when we have no clue what will be inside.
Second : I have suggestion to ease the API of the connector. Since what the connector needs is to retrieve an access_token using the auth_flow_id, maybe it's easier to just pass to a connector the method that does this, and then the connector will be able use it when needed. What do you think about this ?

testinnplayin:
I put in generic terms because I'm thinking of the future of the connectors. We could do similar things with other connectors if we wanted to use this route.

If we have a retry policy in place for the connector, we might need the refresh token in secrets as well and not just the access token.

I'm not sure about your second point sorry.

PR #219, opened 2020-10-02
Update doc generation script

raphaelvignes:
## Change Summary
The Toucan Connectors Doc generation script was broken, I fixed it by changing the way we retrieve information from the connector classes 

Previously, the command provided in the Readme: PYTHONPATH=. python doc/generate.py MyTypeConnector > doc/mytypeconnector.md returned ModuleNotFound exceptions. It seems that the generate.py script was not kept up to date with changes that occured in Toucan Connectors class.

When running this function: k = getattr(toucan_connectors, sys.argv[1]) the MyTypeConnector was not properly imported because of the \__init__.py definition. 

I changed that by directly importing the required module in generate.py to retrieve the new MyTypeConnector attributes there and generate the doc.
 
I also updated the way the strings retrieved from type_ field definition are cleaned by creating the clean_type_str() function.

## Checklist


doc/generate.py
```diff
@@ -3,23 +3,35 @@
...
-
 import toucan_connectors
-
+from importlib import import_module
 
 def doc_or_empty(klass):
     with suppress(AttributeError):
         return klass.__doc__.strip()
 
+def clean_type_str(field_str):
```
rhuille, line 13: Do we agree that this function does not concern the module importation issue ? Or do I miss something ?

doc/generate.py
```diff
@@ -133,8 +155,10 @@ def generate_all_doc(connectors):
...
     if len(sys.argv) > 1:
-        k = getattr(toucan_connectors, sys.argv[1])
+        con = sys.argv[1]
+        mod = import_module(f'toucan_connectors.{con}.{con}_connector', 'toucan_connectors')
+        k = getattr(mod, f'{snake_to_camel(sys.argv[1])}Connector')
         print(generate(k))
     else:
         generate_all_doc(connectors)
-    generate_summmary(connectors)
+    generate_summmary(connectors)
```
rhuille, line 164: An empty end line is missing

doc/generate.py
```diff
@@ -54,12 +66,13 @@ def generate(klass):
     * `id`: str, required
     * `dataset`: Dataset, required
     """
-    doc = [f'# {klass.type} connector', doc_or_empty(klass), '## Data provider configuration']
```
rhuille, line 57: Updates in this function are not related to the importation issue also right ? If so, maybe you can commit it in an other commit and describe why you are doing these change in the commit message ?

doc/generate.py
```diff
@@ -3,23 +3,35 @@
 import sys
 import os
 from contextlib import suppress
-
 import toucan_connectors
```
rhuille, line 6: You do not need to import it anymore ?

doc/generate.py
```diff
@@ -123,7 +144,8 @@ def generate_summmary(connectors):
 
 def generate_all_doc(connectors):
     for key, value in connectors.items():
-        k = getattr(toucan_connectors, key)
+        mod = import_module(f'toucan_connectors.{value}.{value}_connector', 'toucan_connectors')
```
rhuille, line 147: :+1:

rhuille (changes requested):
Thanks for updating the description ! 
Maybe I miss something, but there are others changes that are not related to the importation issue. Could you separate the issue and describe why you make the change in the commit message or in the pr description ?

doc/generate.py
raphaelvignes: You're right, I also changed the cleaning of the string retrieved from class definition. 
I'll add that in the change description.

doc/generate.py
raphaelvignes: I removed this import as it is no longer used.

doc/generate.py
raphaelvignes: I did a separate commit and provided clarifications about the change.

rhuille (approved):

PR #228, opened 2020-10-16
Add a get_connector_config_form method to connectors

luc-leonard:
Some connnectors need to be configured when they are instanciated (eg. Postgres needs an hostname, etc.)
Some connectors need to be configured prior to instanciation, because such config is instance wide. eg GoogleSheets2 needs a client_secret, and a client_id. Without those, it is not possible to instanciate a connector.

This PR adds a method get_connector_config_form, optional, at the class level for ToucanConnectors subclasses. it also add a function get_connector_config_form(cls) that dispatches call to  the appropriate class.get_form.

I also separated the config part of the OAuth2Connector in a Pydantic model.

### DIFF
--- a/toucan_connectors/oauth2_connector/oauth2connector.py
+++ b/toucan_connectors/oauth2_connector/oauth2connector.py
@@ -6,7 +6,7 @@ from urllib import parse as url_parse
 
 from authlib.common.security import generate_token
 from authlib.integrations.requests_client import OAuth2Session
-from pydantic import BaseModel
+from pydantic import BaseModel, SecretStr
 
 
 class SecretsKeeper(ABC):
@@ -25,7 +25,7 @@ class SecretsKeeper(ABC):
 
 class OAuth2ConnectorConfig(BaseModel):
     client_id: str
-    client_secret: str
+    client_secret: SecretStr
     redirect_uri: str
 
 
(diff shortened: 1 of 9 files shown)


toucan_connectors/toucan_connector.py
```diff
@@ -43,6 +43,11 @@ def is_oauth2_connector(cls):
     return hasattr(cls, '_auth_flow') and getattr(cls, '_auth_flow') == 'oauth2'
 
 
+def get_form(cls):
```
davinov, line 46: Maybe it's confusing to have the same name for the _connector_ config form (oauth2), and the _connection_ config form.
It can wait for the coming refacto I think

toucan_connectors/oauth2_connector/oauth2connector.py
```diff
@@ -22,6 +23,11 @@ def load(self, key: str) -> Any:
         """
 
 
+class OauthConnectorConfig(BaseModel):
```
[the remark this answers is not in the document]
luc-leonard: Agreed

toucan_connectors/toucan_connector.py
luc-leonard: I changed the name to get_connector_config_form. maybe get_connector_admin_form maybe ?

davinov (approved):
Thanks ! :tada:

toucan_connectors/oauth2_connector/oauth2connector.py
```diff
@@ -25,7 +25,7 @@ def load(self, key: str) -> Any:
 
 class OAuth2ConnectorConfig(BaseModel):
     client_id: str
-    client_secret: str
+    client_secret: SecretStr
```
davinov, line 28: :+1:

davinov (approved):

PR #232, opened 2020-10-20
feat(http connector): Added xml support

raphaelvignes:
This PR adds the support of XML in HttpAPI connector

## Change Summary

The HttpApi connector can be configured to parse XML responses using xmltodict package. 
* A ResponseType field was added to the connector's configuration, options are xml or json
* An XpathSchema was added to the Data Source, in order to receive the xpath string the user might want to use to parse the xml content of the response
* Once converted to a dict the xml response is parsed with the jq filter provided in the jq_filter field

## Checklist

### DIFF
--- a/tests/http_api/test_http_api.py
+++ b/tests/http_api/test_http_api.py
@@ -368,4 +368,6 @@ def test_parse_xml_response(xml_connector, xml_datasource):
             </response>""",
     )
     df = xml_connector.get_df(xml_datasource)
-    assert df.shape == (2, 7)
+    assert df['login'][0] == 'analytica@fakecompany.com'
+    assert df['timeZone'][0] == 'US/Pacific'
+    assert df['id'][1] == '123'
--- a/tests/http_api/test_http_api.py
+++ b/tests/http_api/test_http_api.py
@@ -1,10 +1,35 @@
 import json
+from xml.etree.ElementTree import ParseError
 
 import pytest
 import responses
 
 from toucan_connectors.common import transform_with_jq
 from toucan_connectors.http_api.http_api_connector import Auth, HttpAPIConnector, HttpAPIDataSource
+
+
+@pytest.fixture
+def xml_connector():
+    data_provider = {
+        'name': 'APIreturningXML',
+        'type': 'HttpAPI',
+        'baseroute': 'http://example.com/api/v1',
+        'responsetype': 'xml',
+    }
+    return HttpAPIConnector(**data_provider)
+
+
+@pytest.fixture()
+def xml_datasource():
+    data_source = {
+        'domain': 'testxml',
+        'name': 'XMLAPI',
+        'url': 'foo/xml',
+        'method': 'GET',
+        'xpath': 'output',
+        'filter': '.output.users.user',
+    }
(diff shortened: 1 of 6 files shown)


raphaelvignes:
Tests were OK on the [Flikr API](https://www.flickr.com/services/api/flickr.favorites.getList.html). 
Waiting a real example to confirm.

tests/http_api/test_http_api.py
```diff
@@ -301,3 +326,46 @@ def test_no_top_level_domain():
...
+                    <user id="19" guid="B9ADBCB81AA2F9BAE040307F02092C2E" login="analytica@fakecompany.com" email="analytica@fakecompany.com"
+                    name="Anna Analyzer" roleId="3" timeZone="US/Pacific"/>
+                    <user id="123" guid="AAFF5218D55ABB9234660001BEC117A9" login="randomuser@fakecompany.com" email="randomuser@fakecompany.com"
+                    name="J. Random User" roleId="2" timeZone="US/Pacific"/>
+                </users>
+            </output>
+            </response>""",
+    )
+    df = xml_connector.get_df(xml_datasource)
+    assert df.shape == (2, 7)
```
davinov, line 371: I think you could assert here the expected dataset: it could be nice to see how the tag name, the attributes and their values are turned into columns and values. (It's not as obvious as it is for dictionnaries)

davinov (commented):
Cool! It's nice to have to be able to start selecting stuff with xpath, and then using jq to finish the data shaping. Nice work!

Here is just a small comment to make the expected test result more explicit:

toucan_connectors/common.py
```diff
@@ -198,6 +198,13 @@ def transform_with_jq(data: object, jq_filter: str) -> list:
     '<a href="https://stedolan.github.io/jq/manual/">documentation</a>',
 )
 
+XpathSchema = Field(
```
davinov, line 201: Should it be a StringField?

tests/http_api/test_http_api.py
raphaelvignes: Thanks for the review, I'll enhance that.

toucan_connectors/common.py
raphaelvignes: I was not sure about that, when I saw
```
   filter: str = FilterSchema
``` 
Defined in common.py, I supposed it would need a specific Schema. But I'm fine to use a StringField

raphaelvignes: I was not sure about that, for FilterSchema it's a Field. But I agree that StringField is more relevant.

raphaelvignes: Are you sure this type exists ? 
image

davinov: Sorry, I took the naming from MongoEngine, not Pydantic ^^
In Pydantic, there is a `StrictStr` (https://pydantic-docs.helpmanual.io/usage/types/#strict-types) type, but it's OK to use Field as well.
My apology for this unrelevant comment

davinov (approved):

raphaelvignes:
Test OK with example query from Client: 

image

PR #233, opened 2020-10-20
fix(oauth2): Added metadata to the OAuth2ConnectorConfig. 

luc-leonard:
the redirect_uri field has been tagged as 'not user provided'. This means that even the admin SHOULD NOT fill this field. It should be field automatically by a backend.


rvi008:
I Believe we can merge this PR.

raphaelvignes (approved):
:+1:

PR #236, opened 2020-10-21
Fix/oauth2connector retrieve tokens for aircall needs the client_id and client_secret

luc-leonard:
Aircall needs the client_id and client_secret sent in the body of fetch_tokens. With this fix, this is now the default behavior.
It is also possible to give arbitrary kwargs to OAuth2Connector:retrieve_tokens, they will be passed-through to the body of the request


raphaelvignes (approved):
OK to remove "test_specific_retrieve_token" as the behavior isn't specific to Aircall. This case is already covered in oAuth2Connector tests.

PR #243, opened 2020-10-23
bumps connectors version to 0.43.8

raphaelvignes:
Bumps connectors version to 0.43.8 to onboard the fix on odbc.sh


labynocle (approved):
:+1:

PR #248, opened 2020-10-28
feat(odbc-connector): enhanced odbc connector query field

raphaelvignes:
## Change Summary

A user of the ODBC connector requested that the query field to be the same as in other SQL connectors. 
I improved that by using the sql widget in the pydantic field.

## Checklist


davinov (approved):
:+1:

PR #299, opened 2021-01-19
feat: handle datetime in nosql_apply_parameters_to_query

fspot:
**Before**

```python
nosql_apply_parameters_to_query(
    {'foo': '{{ bar }}'},
    {'bar': datetime(2020, 12, 31)}
)
```

would return

```python
{'foo': '2020-12-31 00:00:00'}
```

(there was a forced cast to string, making it impossible to keep the date type)

**After**

it returns

```python
{'foo': datetime.datetime(2020, 12, 31, 0, 0)}
```

(you can use `bar | string` or `bar.isoformat()` or `bar.strftime(args)` to get a string output)

### DIFF
--- a/tests/test_common.py
+++ b/tests/test_common.py
@@ -1,3 +1,5 @@
+from datetime import datetime, timedelta
+
 import pytest
 from aiohttp import web
 
@@ -142,9 +144,28 @@ def test_apply_parameter_to_query_do_nothing():
             {'code': 'Paris_France', 'domain': 'Test'},
         ),
         ({'code': '{{city}}_{{country}}', 'domain': 'Test'}, None, {'domain': 'Test'}),
+        (
+            {'column': 'date', 'operator': 'eq', 'value': '{{ t + delta }}'},
+            {'t': datetime(2020, 12, 31), 'delta': timedelta(days=1)},
+            {'column': 'date', 'operator': 'eq', 'value': datetime(2021, 1, 1)},
+        ),
+        (
+            {'column': 'date', 'operator': 'eq', 'value': '{{ t.strftime("%d/%m/%Y") }}'},
+            {'t': datetime(2020, 12, 31)},
+            {'column': 'date', 'operator': 'eq', 'value': '31/12/2020'},
+        ),
+        (
+            {'column': 'date', 'operator': 'in', 'value': '{{ allowed_dates }}'},
+            {'allowed_dates': [datetime(2020, 12, 31), datetime(2021, 1, 1)]},
+            {
+                'column': 'date',
+                'operator': 'in',
+                'value': [datetime(2020, 12, 31), datetime(2021, 1, 1)],
+            },
+        ),
     ],
 )
-def test_apply_parameter_to_query(query, params, expected):
+def test_nosql_apply_parameters_to_query(query, params, expected):
     assert nosql_apply_parameters_to_query(query, params) == expected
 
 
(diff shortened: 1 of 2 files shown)


tests/test_common.py
```diff
@@ -142,9 +144,28 @@ def test_apply_parameter_to_query_do_nothing():
...
+            {'t': datetime(2020, 12, 31)},
+            {'column': 'date', 'operator': 'eq', 'value': '31/12/2020'},
+        ),
+        (
+            {'column': 'date', 'operator': 'in', 'value': '{{ allowed_dates }}'},
+            {'allowed_dates': [datetime(2020, 12, 31), datetime(2021, 1, 1)]},
+            {
+                'column': 'date',
+                'operator': 'in',
+                'value': [datetime(2020, 12, 31), datetime(2021, 1, 1)],
```
davinov, line 163: Great for checkboxes variables :+1:

toucan_connectors/common.py
```diff
@@ -71,26 +71,25 @@ def _render_query(query, parameters):
...
         elif isinstance(query, list):
             return [_render_query(elt, parameters) for elt in deepcopy(query)]
-        elif type(query) is str:
+        elif isinstance(query, str):
             if not _has_parameters(query):
                 return query
-            clean_p = deepcopy(parameters)
+
+            # Replace param templating with jinja templating:
+            query = re.sub(RE_PARAM, r'{{ \g<1> }}', query)
```
davinov, line 79: So easier ^^

By the way, I think we should progressively retire this parameter syntax in favor of Jinja one, which is simpler to read and write. Also, it's complex for users to have two possibilities that does the same thing.
Creating an issue to suggest that: https://github.com/ToucanToco/toucan-connectors/issues/303

toucan_connectors/common.py
```diff
@@ -71,26 +71,25 @@ def _render_query(query, parameters):
...
-                res = ast.literal_eval(res)
-                if isinstance(res, str):
-                    return res
-                else:
-                    return _prepare_result(res)
-            except (SyntaxError, ValueError):
-                return res
+            env = NativeEnvironment()
+            res = env.from_string(query).render(clean_p)
+            # NativeEnvironment's render() isn't recursive, so we need to
```
davinov, line 88: It's not instantly clear to me what do you mean by "render is not recursive".
May you explain that with an example? Thx

(I tried this to test nested arrays, but obtained as expected:
```py
>>> env.from_string('{{ [punk, [ipa]] }}').render({'punk': ['pale', 'ale'], 'ipa': 10})
[['pale', 'ale'], [10]]
```
)

davinov (commented):
Great! Here is one small question for something I've trouble understanding :

toucan_connectors/common.py
fspot: Sure !
There are actually 2 problems : 
The 1st one is that `render` will apply a `ast.literal_eval` on the result (if the result is a `str`), and by doing so we lose the original type : 

```python
>>> env.from_string('{{ x }}').render({'x': 1})  # we expect 1
1
>>> env.from_string('{{ x }}').render({'x': '1'})  # we expect '1'
1  # <- meh
```

to handle this problem, a solution is to put strings inside quotes, this is what `_prepare_parameters` does.

```python
>>> env.from_string('{{ x }}').render({'x': '"1"'})  # we expect '1'
'1'  # <- it works !
```

it also works if x is a list and we get one element from this list : 

```python
>>> env.from_string('{{ x[0] }}').render({'x': ['"1"', '"2"']})  # we expect '1'
'1'
```

but here's the 2nd problem : 

```python
>>> env.from_string('{{ x }}').render({'x': ['"1"', '"2"']})  # we expect ['1', '2']
['"1"', '"2"']  # <- meh
```

in this case, since the result is a `list` and not a `str`, jinja did not apply the `literal_eval` recursively. That's why we have to do it by hand, by calling `_prepare_result`.

luc-leonard (approved):
LGTM

PR #300, opened 2021-01-19
fix(mongo): apply match regex at the end of the pipeline

aklom:
## Change Summary
Adding the match regex to the existing `$match` of the query does not work if the `$match` comes before other stages that affect the field we are filtering on. 

*Example*
We rename the column of the queried field of our dataset. Applying the match with the new field name before renaming the column would result in an empty dataframe, since it will try to find the new field name in the old dataset (before renaming the column) . 
 
We now append the match regex at the end of the query pipeline, that way we are sure any dependent pipeline step will be executed "before" the match regex. 

## Checklist


raphaelvignes:
Thx for this fix !

raphaelvignes (approved):
Thx for this fix !

PrettyWood (approved):
Excellent 👌

PR #316, opened 2021-02-02
fix: support regex with non-string values

PrettyWood:
avoid error `$regexMatch needs 'input' to be of type string`


JeremyPinhel (approved):
nice 👍

PR #322, opened 2021-02-09
fix(GithubConnector): using a get instead of a key on latest_retrieve…

raphaelvignes:
## Change Summary
in github_connector.py, 
changed 
`self.get_pages(
                name=name,
                client=client,
                dataset=dataset,
                organization=organization,
                page_limit=page_limit,
                latest_retrieved_object=latest_retrieved_object[name]
                if latest_retrieved_object
                else None,
            )`
to
`
self.get_pages(
                name=name,
                client=client,
                dataset=dataset,
                organization=organization,
                page_limit=page_limit,
                latest_retrieved_object=latest_retrieved_object.get(name)
                if latest_retrieved_object
                else None,
            )
`
To avoid a KeyError in case the name cannot be found in dict.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -67,7 +67,7 @@ def get_static_file_paths():
 
 setup(
     name='toucan_connectors',
-    version='0.45.7',
+    version='0.45.8',
     description='Toucan Toco Connectors',
     long_description=(HERE / 'README.md').read_text(encoding='utf-8'),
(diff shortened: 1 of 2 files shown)


hachichaud (approved):

raphaelvignes:
Thx for the review.

PR #338, opened 2021-03-05
fix(mongo-connector): make regex match case insensitive

aklom:
## Change Summary
The regex match was case sensitive (making the lazy requester not work properly).


PrettyWood (approved):
👍

PR #347, opened 2021-03-15
feat: clickhouse connector

raphaelvignes:
## Change Summary

New ClickHouse Connector

## Checklist


toucan_connectors/clickhouse/clickhouse_connector.py
```diff
@@ -0,0 +1,117 @@
...
+                        f"""SELECT name FROM system.tables WHERE database = '{current_config["database"]}'"""
+                    )
+                    res = cursor.fetchall()
+                    available_tables = [table[0] for table in res]
+                    constraints['table'] = strlist_to_enum('table', available_tables)
+
+        return create_model('FormSchema', **constraints, __base__=cls).schema()
+
+
+class ClickhouseConnector(ToucanConnector):
```
piotch, line 84: @raphaelvignes I think we should add a configuration option to support SSL wrapped TCP connection ("secure connection"). The protocol part of the url is not `clickhouse` in this case but `clickhouses` (with a S)

https://clickhouse-driver.readthedocs.io/en/latest/api.html#clickhouse_driver.Client.from_url

raphaelvignes: Thanks, I added this option.

toucan_connectors/clickhouse/clickhouse_connector.py
```diff
@@ -0,0 +1,117 @@
...
+    )
+    port: int = Field(None, description='The listening port of your database server')
+    user: str = Field(..., description='Your login username')
+    password: SecretStr = Field('', description='Your login password')
+    ssl_connection: bool = Field(False, description='Create a SSL wrapped TCP connection')
+
+    def get_connection_url(self, *, database='default'):
+        if self.ssl_connection:
+            return f'clickhouses://{self.user}:{self.password.get_secret_value() if self.password else ""}@{self.host}:{self.port}/{database}'
+        return f'clickhouse://{self.user}:{self.password.get_secret_value() if self.password else ""}@{self.host}:{self.port}/{database}'
```
piotch, line 100:
```suggestion
        proto = 'clickhouses' if self.ssl_connection else 'clickhouse'
        return f'{proto}://{self.user}:{self.password.get_secret_value() if self.password else ""}@{self.host}:{self.port}/{database}'
```

piotch (approved):

toucan_connectors/clickhouse/clickhouse_connector.py
raphaelvignes: Thx

Issue #303, opened 2021-01-20
Deprecate %s(..) parameter syntax in favor of Jinja 


davinov:
We actually support two syntax for variables that should be substituted in queries.
This is probably one too much :p
I suggest keeping Jinja only for clarity.

See this code where we already replace the param syntax by jinja: https://github.com/ToucanToco/toucan-connectors/pull/299/files#diff-0f79225d8a2288bbc6507d1647f4e5594e61b8284ca99c89759d678d7f3d3e87R79

hachichaud:
Points to take into consideration :

The syntaxes used in connectors out of our own implementation with Jinka is the one described here https://www.python.org/dev/peps/pep-0249/#paramstyle . It's what you'd find if you check the documentation of some connectors. (For instance, in Snowflake Connector, you find the pep 249 syntaxes described).

Otherwise using jinja syntax offers the advantage of being able to use nested params + we could have only one way of applying parameters.

PR #350, opened 2021-03-17
fix(snowflake): convert parameters to qmark style

hachichaud:
## Change Summary
The parameters value passed from the backend can be incompatible
with pyformat mapping (dict or array values).
I use convert to qmark method that is already used on other connectors
Another way would be to flatten parameters, but it could still leave
some incompatible values (arrays of dict).

## Checklist

### DIFF
--- a/tests/snowflake/test_snowflake.py
+++ b/tests/snowflake/test_snowflake.py
@@ -36,7 +36,7 @@ sd = SnowflakeDataSource(
     domain='test_domain',
     database='test_database',
     warehouse='test_warehouse',
-    query='test_query with %(foo)s and {{ pokemon }}',
+    query='test_query with %(foo)s and %(pokemon)s',
     parameters={'foo': 'bar', 'pokemon': 'pikachu'},
 )
 
@@ -55,7 +55,7 @@ def test_snowflake(mocker):
     sc.get_df(sd)
 
     expected_calls = [
-        call('test_query with %(foo)s and {{ pokemon }}', {'foo': 'bar', 'pokemon': 'pikachu'}),
+        call('test_query with ? and ?', ['bar', 'pikachu']),
     ]
 
     mock_execute = snock.return_value.cursor.return_value.execute
@@ -153,7 +153,7 @@ def test_snowflake_data_source_default_warehouse(mocker):
 
     sc.get_df(ds)
 
-    expected_calls = [call('foo', None)]
+    expected_calls = [call('foo', [])]
 
     snow_mock.return_value.cursor.return_value.execute.assert_has_calls(expected_calls)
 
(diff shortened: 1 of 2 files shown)


raphaelvignes:
Thanks for this fix :+1:

raphaelvignes (approved):

tests/snowflake/test_snowflake.py
```diff
@@ -36,7 +36,7 @@
     domain='test_domain',
     database='test_database',
     warehouse='test_warehouse',
-    query='test_query with %(foo)s and {{ pokemon }}',
+    query='test_query with %(foo)s and %(pokemon)s',
```
davinov, line 39: `%(...)` is an old python syntax, I think we'd prefer to push Jinja' one in our connectors configs
Today, we're already replacing these automatically: https://github.com/ToucanToco/toucan-connectors/blob/ce8ed0622daf262abcec7f72bf0d90f6506c6962/toucan_connectors/common.py#L78

So my suggestion would be to use jinja {{ }} in your tests and documentation as well

davinov (dismissed):
Nice :+1: 
A very small suggestion regarding delimiters used in examples:

tests/snowflake/test_snowflake.py
hachichaud: les syntaxes compatibles avec execute de snowflake sont

* pyformat (`%()s`)  => bug car on passe un dictionnaire parameters qui peut comporter des valeurs invalides.
* qmark => cest dans ce format qu'on convertit notre query pour bypasser le probleme du dictionnaire.

Par contre on ne peut pas utiliser jinja car ce n'est pas safe sur des requetes SQL
cf le code que tu pointes, qui est extrait de la methode no_sql_apply_parameters_to_query :

```
def nosql_apply_parameters_to_query(query, parameters, handle_errors=False):
    """
    WARNING : DO NOT USE THIS WITH VARIANTS OF SQL
    Instead use your client library parameter substitution method.
    https://www.owasp.org/index.php/Query_Parameterization_Cheat_Sheet
    """
```

on veut spécifiquement éviter le templating de cette manière.

hachichaud:
[As a user i can use Snowflake live data queries on a chart or tile](https://trello.com/c/V84rCXYn/442-as-a-user-i-can-use-snowflake-live-data-queries-on-a-chart-or-tile)

davinov (approved):
We had a slack discussion about which style of param to use, let's open an issue/think more about it, but this PR is not the place for that.

tests/snowflake/test_snowflake.py
hachichaud: @davinov you already had an issue on that actually! https://github.com/ToucanToco/toucan-connectors/issues/303
^^

PR #357, opened 2021-03-26
fix: pin python snowflake connector version

kadjou1:
This commit pins the Snowflake python connector to 2.4.1 and removes the previous pin on pyarrow.


kadjou1:
[(3) Pin la version de snowflake connector dans le repo connector](https://trello.com/c/Hv7F5L0y/401-3-pin-la-version-de-snowflake-connector-dans-le-repo-connector)

hachichaud (approved):

PR #369, opened 2021-04-12
refactor(sql connectors): add util pandas_read_sql

fspot:
## Change Summary

Centralize sql query manipulations in a dedicated util, so it's easier to add or modify common behaviours.

## Checklist

### DIFF
--- a/toucan_connectors/common.py
+++ b/toucan_connectors/common.py
@@ -310,6 +310,17 @@ def adapt_param_type(params):
     return {k: (tuple(v) if isinstance(v, list) else v) for (k, v) in params.items()}
 
 
+def extract_table_name(query: str) -> str:
+    m = re.search(r'from\s*(?P<table>[^\s;]+)\s*(where|order by|group by|limit)?', query, re.I)
+    table = m.group('table')
+    return table
+
+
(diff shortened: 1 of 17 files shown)


raphaelvignes (approved):
Merci pour cette refacto, ça rend les connecteurs SQL beaucoup plus propres !

toucan_connectors/common.py
```diff
@@ -307,3 +308,26 @@ def adapt_param_type(params):
...
+    **kwargs,
+) -> pd.DataFrame:
+    if convert_to_printf:
+        query = convert_to_printf_templating_style(query)
+    if render_user:
+        query = Template(query).render({'user': params.get('user', {})})
+    if convert_to_qmark:
+        query, params = convert_to_qmark_paramstyle(query, params)
+    if adapt_params:
+        params = adapt_param_type(params)
```
davinov, line 330: I'm bit dubious of factorization where we introduce a set of conditional branches. I also find that the the function name doesn't convey the sense of this operations.
But this is not blocking for now, as it's fairly easy to go back and fourth between implementations for now, and I don't have an alternative to suggest ^^

davinov (approved):

PR #371, opened 2021-04-13
fix(nosql apply parameters to query): keep str type when appropriate

fspot:
## Change Summary

In some cases, we used to lose the `str` type when using the util nosql_apply_parameters_to_query, leading to errors.

## Checklist

### DIFF
--- a/toucan_connectors/common.py
+++ b/toucan_connectors/common.py
@@ -32,6 +32,24 @@ class NonValidVariable(Exception):
     """ Error thrown for a non valid variable in endpoint """
 
 
+def is_jinja_alone(s: str) -> bool:
+    """
+    Return True if the given string is a jinja template alone.
+
+    For example, these strings are jinja templates alone:
+        '{{ foo }}'
+        '{{ foo + bar }}'
+        '{%if my_indic%}1{%else%}2{%endif%}'
+
+    Whereas these strings are not:
+        'Hey {{ foo }} !'
+        '{{ foo }}{{ bar }}'
+
+    In the 2nd case, we will always render the result as a string.
+    """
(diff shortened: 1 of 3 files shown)


toucan_connectors/common.py
```diff
@@ -32,6 +32,24 @@ class NonValidVariable(Exception):
...
 
 
+def is_jinja_alone(s: str) -> bool:
+    """
+    Return True if the given string is a jinja template alone.
+
+    For example, these strings are jinja templates alone:
+        '{{ foo }}'
+        '{{ foo + bar }}'
+        '{%if my_indic%}1{%else%}2{%endif%}'
```
davinov, line 42: Curiosity: do we have cases where we use such syntax?

Also it's difficult to say it's just jinja alone if it starts and end by {% %}. An example would be:
`'{%if count %}{{ count }}{%else%}No{%endif%} chair{% if count > 1 %}s{% endif %}`'. In that case, keeping a string would be necessary.

davinov (commented):
Very clear, thanks for the detailed doc string and associated tests :+1: 
One question:

toucan_connectors/common.py
fspot: On a surtout des syntaxes de type `"{{ ... }}"`. La 2e syntaxe je pense pas qu'on en ai des masses. Par contre après vérification le NativeEnvironment jinja gère bien ce cas aussi en fait, donc je vais juste ajouter des tests pour ça.

fspot: Ajouté dans 93a64e8

davinov: Excellent that NativeEnv handles it, thanks for the new tests!

davinov (approved):

PR #426, opened 2021-06-10
Azure mssql parametrization

fspot:
Parametrization was never implemented on this connector :sweat: let's do it


luc-leonard (approved):
Merci pour cette PR

PR #434, opened 2021-06-22
fix(hana): support paramters

piotch:


piotch:
Closed because https://github.com/ToucanToco/toucan-connectors/issues/440

PR #466, opened 2021-08-09
feat(get_identifier): use method to get an unique identifier

f-bb-toucantoco:
## Change Summary
Use new method to get_identifier (method defined for cache)
- Implement a new method to get_identifier from Connector information
- Implement specific method for SnowflakeoAuth2Connector (impossible to serialize object from OAuth2Connector)

## Checklist

### DIFF
--- a/toucan_connectors/snowflake_oauth2/snowflake_oauth2_connector.py
+++ b/toucan_connectors/snowflake_oauth2/snowflake_oauth2_connector.py
@@ -1,4 +1,5 @@
 import logging
+import uuid
 from contextlib import suppress
 from timeit import default_timer as timer
 from typing import Any, Dict, List, Optional
@@ -9,6 +10,7 @@ from pydantic import Field, SecretStr, create_model
 from snowflake.connector import SnowflakeConnection
 
 from toucan_connectors.connection_manager import ConnectionManager
+from toucan_connectors.json_wrapper import JsonWrapper
 from toucan_connectors.oauth2_connector.oauth2connector import (
     OAuth2Connector,
     OAuth2ConnectorConfig,
@@ -194,7 +196,7 @@ class SnowflakeoAuth2Connector(ToucanConnector):
                     raise TypeError('close is not a function')
 
         connection: SnowflakeConnection = connection_manager.get(
-            self.identifier,
+            identifier=self.get_identifier(),
             connect_method=connect_function,
             alive_method=alive_function,
             close_method=close_function,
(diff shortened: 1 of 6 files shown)


f-bb-toucantoco:
[(3) [Snowflake - identity] - update class ConnectionManager to use the new method to get a key](https://trello.com/c/3v4Ik6Bg/1083-3-snowflake-identity-update-class-connectionmanager-to-use-the-new-method-to-get-a-key)

toucan_connectors/snowflake_oauth2/snowflake_oauth2_connector.py
```diff
@@ -203,6 +205,20 @@ def close_function(conn: SnowflakeConnection) -> None:
...
+                'name': self.name,
+                'account': self.account,
+                'client_id': self.client_id,
+                'scope': self.scope,
+                'role': self.role,
+            },
+            sort_keys=True,
+        )
+        string_uid = str(uuid.uuid3(uuid.NAMESPACE_OID, json_uid))
+        return string_uid
```
Sanix-Darker, line 220: The idea for me here is good... but instead of passing the whole json string object to `uuid3`, i suggest to pass only a hash of the json string of the object, because following the [documentation](https://docs.python.org/3/library/uuid.html#uuid.uuid3) it should be only a `name` not a huge json string with a lot of elements in it, so am proposing this:

```python
# at the top of the file i import a builtin function
# md5 from hashlib / instead of uuid
from hashlib import md5

# ... some code here ...

# and i will structure the function in this way...
def get_identifier(self) -> str:
    # we get the hash of the json string
    json_uid = md5(JsonWrapper.dumps(
        {
            'name': self.name,
            'account': self.account,
            'client_id': self.client_id,
            'scope': self.scope,
            'role': self.role,
        },
        sort_keys=True,
    ).encode("utf-8")).hexdigest()
    
    return str(uuid.uuid3(uuid.NAMESPACE_OID, json_uid))
```

toucan_connectors/toucan_connector.py
```diff
@@ -432,7 +432,12 @@ def get_cache_key(
             del data_source_rendered['parameters']
             unique_identifier['datasource'] = data_source_rendered
 
-        json_uid = json.dumps(unique_identifier, sort_keys=True)
+        json_uid = JsonWrapper.dumps(unique_identifier, sort_keys=True)
+        string_uid = str(uuid.uuid3(uuid.NAMESPACE_OID, json_uid))
```
Sanix-Darker, line 436: Same principe for `hash` here, as a proposition...

PR #467, opened 2021-08-09
feat(optimize_connection_manager): check if connection use good database and/or warehouse

f-bb-toucantoco:
## Change Summary
Check if the connection use the database or warehouse in DataSource
If not, execute a query to change it

## Checklist

### DIFF
--- a/toucan_connectors/snowflake_common.py
+++ b/toucan_connectors/snowflake_common.py
@@ -5,7 +5,7 @@ from typing import Dict, List, Optional
 
 import pandas as pd
 from pydantic import Field, constr
-from snowflake.connector import DictCursor
+from snowflake.connector import DictCursor, SnowflakeConnection
 
 from toucan_connectors.query_manager import QueryManager
 from toucan_connectors.sql_query_helper import SqlQueryHelper
@@ -81,7 +81,7 @@ class SnowflakeCommon:
 
     def _execute_query_internal(
         self,
-        connection,
+        connection: SnowflakeConnection,
         query: str,
         query_parameters: Optional[dict] = None,
     ) -> pd.DataFrame:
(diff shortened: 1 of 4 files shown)


f-bb-toucantoco:
[(3) [Snowflake - identity] - optimize method to define where is the connection](https://trello.com/c/EmN5l7zO/1151-3-snowflake-identity-optimize-method-to-define-where-is-the-connection)

toucan_connectors/snowflake_common.py
```diff
@@ -171,28 +171,36 @@ def _execute_parallelized_queries(
...
     ) -> pd.DataFrame:
-        self._execute_query(connection, f'USE DATABASE {data_source.database}')
-        self._execute_query(connection, f'USE WAREHOUSE {data_source.warehouse}')
+        prepare_query: List = []
+        if data_source.database != connection.database:
+            prepare_query.append(f'USE DATABASE {data_source.database}')
+        if data_source.warehouse != connection.warehouse:
+            prepare_query.append(f'USE WAREHOUSE {data_source.warehouse}')
+        if len(prepare_query) > 0:
+            self._execute_query(connection, ';'.join(prepare_query) + ';')
```
Sanix-Darker, line 186: Nice, just a proposition, of a short way, with ternaries... 
Without using list... :
```python
self._execute_query(connection, 
    (f'USE DATABASE {data_source.database};' if data_source.database != connection.database else '') + \
    (f'USE WAREHOUSE {data_source.warehouse};' if data_source.warehouse != connection.warehouse else '')
);
```

f-bb-toucantoco: What is the advantage of this solution over using a list?

Sanix-Darker: Actually, there is not a huge advantage with this approach... but we have : 
- less lines of code.
- no variables (it's not a big deal, but the creation of a list, the append on it and the join on it can be small in term of resources for one call in dev.. but for thousands of calls in prod... )
- if we're going to add more prepared queries in the future, we will prevent manipulating variables(append/join) and directly concatenate the final query depending on conditions... with less lines.


i assume it's not a big deal after all 😅 , but in terms of  `'optimizations'` it can be nice.... i guess.... ?

PR #468, opened 2021-08-11
feat(google_bigquery_new): new method to retrieve data in bigquery with pandas

f-bb-toucantoco:
## Change Summary
Remove pandas_gbq to replace by google.bigquery dependency

With pandas_gbq we have a security breach and we need to remove it

## Checklist

### DIFF
--- a/toucan_connectors/google_big_query/google_big_query_connector.py
+++ b/toucan_connectors/google_big_query/google_big_query_connector.py
@@ -1,13 +1,13 @@
 import logging
 from enum import Enum
 from timeit import default_timer as timer
-from typing import List
+from typing import Dict, List
 
+import pandas
 import pandas as pd
-import pandas_gbq
+from google.cloud import bigquery
 from pydantic import Field
 
-from toucan_connectors.common import apply_query_parameters
 from toucan_connectors.google_credentials import GoogleCredentials, get_google_oauth2_credentials
 from toucan_connectors.toucan_connector import ToucanConnector, ToucanDataSource
 
@@ -24,6 +24,20 @@ class GoogleBigQueryDataSource(ToucanDataSource):
         '<a href="https://cloud.google.com/bigquery/docs/reference/standard-sql/query-syntax">here</a>',
         widget='sql',
     )
+
+
+# TODO - check bad return type
+def _define_type(value) -> str:
+    if isinstance(value, bool):
+        return 'BOOL'
+    elif isinstance(value, int):
+        return 'NUMERIC'
+    elif isinstance(value, float):
+        return 'FLOAT64'
+    elif isinstance(value, str):
+        return 'STRING'
+    else:
+        return 'STRING'
 
 
 class GoogleBigQueryConnector(ToucanConnector):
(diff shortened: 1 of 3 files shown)


f-bb-toucantoco:
[(3) Connecteur big query allows SQL injection](https://trello.com/c/K900uaXC/1054-3-connecteur-big-query-allows-sql-injection)

toucan_connectors/google_big_query/google_big_query_connector.py
```diff
@@ -53,24 +67,39 @@ class GoogleBigQueryConnector(ToucanConnector):
...
-            credentials=credentials,
-            dialect=self.dialect,
+        result = (
+            client.query(
+                query, job_config=bigquery.QueryJobConfig(query_parameters=parameters)
+            ).result()
+            # Use to generate directly a dataframe pandas
+            .to_dataframe(
+                create_bqstorage_client=True,
+            )
```
Sanix-Darker, line 102: Would have suggest : 
```python
).result()
.to_dataframe(
    create_bqstorage_client=True,
) # Use to generate directly a dataframe pandas
```
just to prevent putting comments between object's function calls...

toucan_connectors/google_big_query/google_big_query_connector.py
```diff
@@ -53,24 +67,39 @@ class GoogleBigQueryConnector(ToucanConnector):
...
-        """
-        Uses Pandas read_gbq method to extract data from Big Query into a dataframe
-        See: http://pandas.pydata.org/pandas-docs/stable/generated/pandas.io.gbq.read_gbq.html
-        Note:
-            The parameter reauth is set to True to force Google BigQuery to reauthenticate the user
-            for each query. This is necessary when extracting multiple data to avoid the error:
-            [Errno 54] Connection reset by peer
-        """
+    @staticmethod
+    def _get_google_credentials(credentials, scopes):
```
Sanix-Darker, line 71: credentials type ?
scope type ?

toucan_connectors/google_big_query/google_big_query_connector.py
```diff
@@ -53,24 +67,39 @@ class GoogleBigQueryConnector(ToucanConnector):
...
+        credentials = get_google_oauth2_credentials(credentials).with_scopes(scopes)
+        return credentials
+
+    @staticmethod
+    def _connect(credentials) -> bigquery.Client:
```
Sanix-Darker, line 76: credentials type ?

toucan_connectors/google_big_query/google_big_query_connector.py
```diff
@@ -53,24 +67,39 @@ class GoogleBigQueryConnector(ToucanConnector):
...
+                    'operation': 'connect',
+                    'execution_time': end - start,
+                    'connector': 'google_big_query',
+                }
+            },
+        )
+        return client
+
+    @staticmethod
+    def _execute_query(client: bigquery.Client, query: str, parameters: List) -> pandas.DataFrame:
```
Sanix-Darker, line 93: it's a security check but i was thinking it will be great to have a `try - catch` here, so we can check either or not the query was well perform... because we never knows when communicating with external systems...

maybe this is overdoing too... but am just suggesting...

toucan_connectors/google_big_query/google_big_query_connector.py
```diff
@@ -84,3 +113,36 @@ def _retrieve_data(self, data_source: GoogleBigQueryDataSource) -> pd.DataFrame:
...
+            f'Play request {data_source.query} with parameters {data_source.parameters}'
+        )
+
+        credentials = GoogleBigQueryConnector._get_google_credentials(self.credentials, self.scopes)
+        query = GoogleBigQueryConnector._prepare_query(data_source.query)
+        parameters = GoogleBigQueryConnector._prepare_parameters(query, data_source.parameters)
+
+        client = GoogleBigQueryConnector._connect(credentials)
+        result = GoogleBigQueryConnector._execute_query(client, query, parameters)
+        # client.close()
```
Sanix-Darker, line 146: when are we closing the client's session... if it's not there ?

toucan_connectors/google_big_query/google_big_query_connector.py
```diff
@@ -84,3 +113,36 @@ def _retrieve_data(self, data_source: GoogleBigQueryDataSource) -> pd.DataFrame:
...
+                query_parameters.append(
+                    bigquery.ScalarQueryParameter(k, _define_type(parameters[k]), parameters[k])
+                )
+        return query_parameters
+
+    @staticmethod
+    def _prepare_query(query: str) -> str:
+        # replace ToucanToco variable definition by Google Big Query variable definition
+        new_query = query.replace('{{', '@').replace('}}', '')
+        return new_query
```
Sanix-Darker, line 133: because it's just a replacement... is it possible to just return the final result instead of creating a variable for only one instruction 🤔 ? like : 
```python
return query.replace('{{', '@').replace('}}', '')
```

in a big overwiew... if we are sure this is the only operation we're doing... do we need a whole function to store that operation ? or maybe am wrong... and others steps are coming ? 🤔

Sanix-Darker (approved):

PR #472, opened 2021-08-17
refactor(oauth2 connectors): the _oauth2_connector property of oauth2 connectors is no longer set using __dict__

luc-leonard:
## Change Summary
refactor(oauth2 connectors): the _oauth2_connector property of oauth2 connectors is no longer set using __dict__. It is now properly declared as private variables for pydantic. This excludes them from the json generated by `.json()`


f-bb-toucantoco (approved):

davinov (approved):
Excellent idea!
Note for later: maybe we can even add the `Config.underscore_attrs_are_private` option to ensure this will be by default

PR #493, opened 2021-09-27
feat(dataslice): add a arbitrary dict to store metadatas in dataslices.

luc-leonard:
## Change Summary
add a arbitrary dict to store metadatas in dataslices.

### DIFF
--- a/toucan_connectors/toucan_connector.py
+++ b/toucan_connectors/toucan_connector.py
@@ -32,7 +32,7 @@ class DataStats(NamedTuple):
     execution_time: Optional[float] = None  # query's execution time in ms
     conversion_time: Optional[float] = None  # Result conversion to DataFrame time
     df_memory_size: Optional[int] = None  # Dataframe's memory usage in bytes
-    metadata: Dict[str, Any] = {}
+    others: Dict[str, Any] = {}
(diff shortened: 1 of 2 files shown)


toucan_connectors/toucan_connector.py
```diff
@@ -32,6 +32,7 @@ class DataStats(NamedTuple):
     execution_time: Optional[float] = None  # query's execution time in ms
     conversion_time: Optional[float] = None  # Result conversion to DataFrame time
     df_memory_size: Optional[int] = None  # Dataframe's memory usage in bytes
+    metadata: Dict[str, Any] = {}
```
davinov, line 35: Maybe we should just name that `others`? So it's accessible using slice.stats.others
I'm starting to fear that we use the term `meta` in a lot of places with different meanings. As it doesn't mean something in its own, it could be confusing

davinov: (or something else if you have a better idea)

luc-leonard: this is a good remark, `meta` is used in a lot of place in the codebase.

other is fine

davinov (approved):

PR #495, opened 2021-10-04
fix(all): cache key with types not json serializable

davinov:
Like dates (mongo queries, for instance, can contain some of there types)
It's OK to use here something that would not be the exact same when re-parsing, as we never re-parse the query cache key (only used to be hashed.

The only edge case I saw using a `default=str` is that if two query could have the same key if someone uses the exact string representation of a date, and then the date (or vice-versa). That could actually happen quite often of someone is unsure of the type, and tries both in sequence. So i went for the `hash` function instead.


tests/mongo/test_mongo.py
```diff
@@ -683,3 +684,26 @@ def test_get_cache_key_with_datasource(mongo_connector, mongo_datasource):
...
+    cache_key_date = mongo_connector.get_cache_key(datasource_with_date)
+
+    datasource_with_str = mongo_datasource(
+        collection='test_col',
+        query={'domain': 'plop', 'date': str(datetime(2021, 10, 4))},
+        parameters={'DOMAIN': 'domain1', 'FOO': 'BAR'},
+    )
+    cache_key_str = mongo_connector.get_cache_key(datasource_with_str)
+
+    assert cache_key_str != cache_key_date
```
austil, line 709: On veut vérifier quoi avec cette assertion ? Que la date est sérialisée avec autre chose que `str` ?

davinov: A query with a stringified date and query with an actual date should not yield the same cache key.
This is because there two queries will lead to different results in mongo: `find({ date: ISODate(2021, 10, 4)})` and `find({ date: '2021-10-04 00:00:00})`.

toucan_connectors/toucan_connector.py
```diff
@@ -446,7 +446,7 @@ def get_cache_key(
         if data_source is not None:
             unique_identifier['datasource'] = self._render_datasource(data_source)
 
-        json_uid = JsonWrapper.dumps(unique_identifier, sort_keys=True)
+        json_uid = JsonWrapper.dumps(unique_identifier, sort_keys=True, default=hash)
```
luc-leonard, line 449: Clever :+1:

luc-leonard (approved):

PR #500, opened 2021-10-18
fix(snowflake): invalid token does not prevent to retrieve a new one

davinov:
Since https://github.com/ToucanToco/toucan-connectors/pull/497, some claims in the token are validated (even if there is `'verify_signature': False`), which could prevent new token to be fetched. There was then no way to connect, once the validity has been broken once.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -71,7 +71,7 @@ def get_static_file_paths():
 
 setup(
     name='toucan_connectors',
-    version='1.3.19',
+    version='1.3.20',
     description='Toucan Toco Connectors',
     long_description=(HERE / 'README.md').read_text(encoding='utf-8'),
     long_description_content_type='text/markdown',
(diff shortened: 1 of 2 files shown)


toucan_connectors/snowflake/snowflake_connector.py
```diff
@@ -239,11 +240,24 @@ def get_connection_params(self):
...
-                self.access_token,
-                options={'verify_signature': False},
-            )
-            if datetime.fromtimestamp(access_token['exp']) < datetime.now():
+            try:
+                # Here, we only want to access the expiration date, we don't care if the token is not valid
+                access_token = jwt.decode(
+                    self.access_token,
+                    options={
+                        'verify_signature': False,  # this should be enough according to docs, but doesn't seem like it is
```
raphaelvignes, line 248: I guess this has changed :/

raphaelvignes (approved):
Lint the code and then it's GTM :)

PR #501, opened 2021-10-18
fix(snowflake): require dependency with pyjwt

davinov:
When the jwt support has been added to SnowflakeConector, the dependency has not been added to this connector's requirements.
This means, that, even with a clean install of `toucan-connectors[snowflake]` with pip, the statement `import jwt` could fail, and the connector will not be available.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -71,7 +71,7 @@ def get_static_file_paths():
 
 setup(
     name='toucan_connectors',
-    version='1.3.20',
+    version='1.3.21',
     description='Toucan Toco Connectors',
     long_description=(HERE / 'README.md').read_text(encoding='utf-8'),
     long_description_content_type='text/markdown',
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -35,7 +35,7 @@
     'ROK': ['requests', 'pyjwt', 'simplejson'],
     'sap_hana': ['pyhdb>=0.3.4'],
     'soap': ['zeep', 'lxml==4.2.5'],
-    'snowflake': ['snowflake-connector-python>=2.5'],
+    'snowflake': ['snowflake-connector-python>=2.5', 'pyjwt>=1.6.3'],
```
davinov, line 38: I choose 1.6.3 because it's the first version when the `PyJWTError` exception exist: https://pyjwt.readthedocs.io/en/latest/changelog.html#v1-6-3
(and we import it in the snowflake_connector module)

raphaelvignes (approved):
Thanks for this fix !

PR #502, opened 2021-10-19
fix(deps): remove pyjwt version constraint

davinov:
Some users fo this repo can't upgrade to pyjwt >=1.6.3, so le'ts remove this constraint


Sanix-Darker (approved):
LGTM

PR #507, opened 2021-11-02
feat(snowflake): add a default order by at the end of the query TCTC-659

Sanix-Darker:
## Change Summary
- Added a default Order By at the end of a Snowflake query for consistent pagination
- Cleaned the prepare_limit_query function from sql_query_helper

## Checklist

* [x] Coverage of the code

### DIFF
--- a/toucan_connectors/snowflake_common.py
+++ b/toucan_connectors/snowflake_common.py
@@ -147,13 +147,8 @@ class SnowflakeCommon:
         query: str,
         query_columns: Optional[List] = None,
     ):
-        """
-        if it's a SELECT type query and there is not an 'ORDER BY' inside,
-        we sort with the first column of the dataset
-
-        query: the query sting
-        query_columns: the query's columns
-        """
+        """If it's a SELECT type query and there is not an 'ORDER BY' inside, we sort with the first column of the
+        dataset"""
 
         if query_columns and 'WITH SELECT_STEP_' in query and ' ORDER BY ' not in query:
             # we append the order from row_number on the first column
--- a/toucan_connectors/snowflake_common.py
+++ b/toucan_connectors/snowflake_common.py
@@ -1,6 +1,7 @@
 import concurrent
 import json
 import logging
+from re import search
 from timeit import default_timer as timer
 from typing import Dict, List, Optional
 
@@ -147,12 +148,15 @@ class SnowflakeCommon:
         query: str,
         query_columns: Optional[List] = None,
     ):
-        """If it's a SELECT type query and there is not an 'ORDER BY' inside, we sort with the first column of the
-        dataset"""
+        """If it's a SELECT type query and there is not an 'ORDER BY' inside, we sort using ROW_NUMBER function"""
 
-        if query_columns and 'WITH SELECT_STEP_' in query and ' ORDER BY ' not in query:
+        if (
+            query_columns
+            and search(r'SELECT[\r\n|\n| |\t]', query.upper())
+            and not search(r'ORDER[\r\n|\n| |\t]BY', query.upper())
+        ):
             # we append the order from row_number on the first column
-            query += f' ORDER BY ROW_NUMBER() OVER (ORDER BY {query_columns[0]} ASC)'
+            query += f' ORDER BY {", ".join(query_columns)}'
 
         return query
 
--- a/toucan_connectors/snowflake_common.py
+++ b/toucan_connectors/snowflake_common.py
@@ -142,6 +142,25 @@ class SnowflakeCommon:
             'parameters': prepared_query_parameters,
         }
 
+    def add_default_order_if_needed(
+        self,
+        query: str,
+        query_columns: Optional[List] = None,
+    ):
+        """
+        if it's a SELECT type query and there is not an 'ORDER BY' inside,
+        we sort with the first column of the dataset
+
+        query: the query sting
+        query_columns: the query's columns
+        """
+
+        if query_columns and 'WITH SELECT_STEP_' in query and ' ORDER BY ' not in query:
+            # we append the order from row_number on the first column
+            query += f' ORDER BY ROW_NUMBER() OVER (ORDER BY {query_columns[0]} ASC)'
+
+        return query
+
     def _execute_parallelized_queries(
         self,
         connection: SnowflakeConnection,
@@ -205,8 +224,18 @@ class SnowflakeCommon:
             self.logger.info(f'Connection changed to use  warehouse {connection.warehouse}')
             self._execute_query(connection, f'USE WAREHOUSE {data_source.warehouse}')
 
+        # We add the order if needed
+        data_source.query = self.add_default_order_if_needed(
+            data_source.query, data_source.query_object['columns']
+        )
+
         ds = self._execute_parallelized_queries(
-            connection, data_source.query, data_source.parameters, offset, limit, get_row_count
+            connection,
+            data_source.query,
+            data_source.parameters,
+            offset,
+            limit,
+            get_row_count,
         )
         return ds.df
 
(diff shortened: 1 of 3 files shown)


toucan_connectors/snowflake_common.py
```diff
@@ -142,6 +142,20 @@ def render_datasource(self, datasource: SfDataSource) -> dict:
             'parameters': prepared_query_parameters,
         }
 
+    def add_default_order_if_needed(
+        self,
+        query: str,
+        query_columns: Optional[List] = None,
+    ):
+        """If it's a SELECT type query and there is not an 'ORDER BY' inside, we sort with the first column of the
```
piotch, line 150:
```suggestion
        """If it's a SELECT type query and there is not an 'ORDER BY' inside, we sort using ROW_NUMBER function
```

toucan_connectors/snowflake_common.py
```diff
@@ -142,6 +142,20 @@ def render_datasource(self, datasource: SfDataSource) -> dict:
...
+        dataset"""
+
+        if query_columns and 'WITH SELECT_STEP_' in query and ' ORDER BY ' not in query:
+            # we append the order from row_number on the first column
+            query += f' ORDER BY ROW_NUMBER() OVER (ORDER BY {query_columns[0]} ASC)'
```
piotch, line 155: je comprends pas pourquoi tu as besoin de `ORDER BY {query_columns[0]}` si tu utilise ROW_NUMBER ?

toucan_connectors/snowflake_common.py
Sanix-Darker: updated !

toucan_connectors/snowflake_common.py
Sanix-Darker: Parceque c'est principalement la synthaxe de ROW_NUMBER() qui veut cela !
Peek 2021-11-03 09-24

toucan_connectors/snowflake_common.py
```diff
@@ -142,6 +142,19 @@ def render_datasource(self, datasource: SfDataSource) -> dict:
...
         }
 
+    def add_default_order_if_needed(
+        self,
+        query: str,
+        query_columns: Optional[List] = None,
+    ):
+        """If it's a SELECT type query and there is not an 'ORDER BY' inside, we sort using ROW_NUMBER function"""
+
+        if query_columns and 'WITH SELECT_STEP_' in query and ' ORDER BY ' not in query:
```
davinov, line 152: This may be sensitive to case, where SQL operators should not.
What do you think about doing a case *in*sensistive search? I would use a regexp (`re`) for this task.

toucan_connectors/snowflake_common.py
```diff
@@ -142,6 +142,19 @@ def render_datasource(self, datasource: SfDataSource) -> dict:
...
+            # we append the order from row_number on the first column
+            query += f' ORDER BY ROW_NUMBER() OVER (ORDER BY {query_columns[0]} ASC) ASC'
```
davinov, line 154: Is there any advatage of using the `row_number` function instead of just using the first column?
`ORDER BY {query_columns[0]} ASC`

davinov: Another question about this ordering function: imagine this column does not have unique values, bu contains just a few (like a category). Then this column alone would not be enough to sort the results. To make sure we have always the same order, should we use all columns in the order by clause?

toucan_connectors/sql_query_helper.py
```diff
@@ -32,12 +32,9 @@ def prepare_limit_query(
...
         if not query_check.startswith('show') and not query_check.startswith('describe'):
-            if limit and offset:
-                prepared_query = prepared_query.replace(';', '')
-                prepared_query = f'SELECT * FROM ({prepared_query}) LIMIT {limit} OFFSET {offset};'
-            elif limit:
-                prepared_query = prepared_query.replace(';', '')
-                prepared_query = f'SELECT * FROM ({prepared_query}) LIMIT {limit};'
+            if limit:
+                prepared_query = f'SELECT * FROM ({prepared_query.replace(";", "")}) LIMIT {limit}'
+                prepared_query += f' OFFSET {offset};' if offset else ';'
```
davinov, line 37: The "order by" could be very expensive if wwe do it without a limit. So I would recommend doing the `add_default_order_if_needed` only in this `if`.

davinov (changes requested):
Thanks for working on this! It's not an easy topic ^^
I have a few questions:

toucan_connectors/snowflake_common.py
Sanix-Darker: the `re` search in this case may be too expensive in computing for each query sent to snowflake... is it not better to just uppercase the query and do checks ? 🤔 
```
 if query_columns and 'WITH SELECT_STEP_' in query.upper() and ' ORDER BY ' not in query.upper():
 ```
 
 What do you think ?

File: toucan_connectors/snowflake_common.py (hunk 4 above)
Comment (Sanix-Darker) [reply toucan_connectors/snowflake_common.py:154]: For the first question :
It was the first approach in this PR but we decided with @piotch that it should be better to use an ORDER BY ROWNUMBER function on the first column to prevent ordering on a column that have duplicated values !

For the second question : 
(And yes it's related to the First Question)
This is true, a workaround i propose is this : Instead of orther only on one column, order by all the available columns : 
```
SELECT * FROM (
  WITH SELECT_STEP_0 AS (
    SELECT COLUMN1, COLUMN2, COLUMN3 FROM "ROLE2_SCHEMA"."DEFORESTATION_CSV"
  ) SELECT COLUMN1, COLUMN2, COLUMN3 FROM SELECT_STEP_0 
  ORDER BY ROW_NUMBER() OVER (ORDER BY COLUMN1, COLUMN2, COLUMN3) ASC
) LIMIT 10000;
```

What do you think ?

File: toucan_connectors/snowflake_common.py (hunk 3 above)
Comment (davinov) [reply toucan_connectors/snowflake_common.py:152]: I don't know how expensive would be a regex search, but yes, using upper is already a good idea!
There also could be issues with the spaces before or after (could be line beeaks? or maybe other characters?), but that's very edgy.
So let's go for the upper! :+1:

File: toucan_connectors/snowflake_common.py (hunk 4 above)
Comment (davinov) [reply toucan_connectors/snowflake_common.py:154]: OK! Thanks for the explanation!
1. If we do have the insurance that the result of row_number will stay the same across queries, why not! But I couldn't find anything in the docs indicating it, so maybe we would need to test that.

2. Great proposal :+1:

File: toucan_connectors/snowflake_common.py (hunk 3 above)
Comment (Sanix-Darker) [reply toucan_connectors/snowflake_common.py]: You marked a point for the line breaks between ORDER and BY so finally i used the regex search on that !

File: toucan_connectors/snowflake_common.py (hunk 4 above)
Comment (Sanix-Darker) [reply toucan_connectors/snowflake_common.py]: updated !

File: toucan_connectors/sql_query_helper.py (hunk 5 above)
Comment (Sanix-Darker) [reply toucan_connectors/sql_query_helper.py:37]: Just to be sure, at the end of our discussion we decided to let this like that right ? 🤔

File: toucan_connectors/snowflake_common.py [hunk 6]
```diff
@@ -147,12 +148,15 @@ def add_default_order_if_needed(
...
         query_columns: Optional[List] = None,
     ):
-        """If it's a SELECT type query and there is not an 'ORDER BY' inside, we sort with the first column of the
-        dataset"""
+        """If it's a SELECT type query and there is not an 'ORDER BY' inside, we sort using ROW_NUMBER function"""
 
-        if query_columns and 'WITH SELECT_STEP_' in query and ' ORDER BY ' not in query:
+        if (
+            query_columns
+            and 'SELECT' in query.upper()
```
davinov, line 155: :+1: Great, thanks!
Maybe we should also update the tests so that they contain SELECT and not WITH SELECT_STEP?

Sanix-Darker: Updated !

piotch:
Closed because we are not able to provide this guarantee in place of snowflake at the moment.

PR #508, opened 2021-11-09
feat(http): support for custom auth-schemes when using custom_token_server

piotch:
## Change Summary

RFC 2617 standard on HTTP Authentication states that credentials should consist of at least a scheme and arbitrary parameters:

`credentials = auth-scheme #auth-param`

In practice auth-scheme is often "Bearer" because it is used in the OAuth2 standard [1, 2]. This is our default, the initial request made to the `custom_token_server` will be used in the the Authorization header after the Bearer auth-scheme.

This PR adds support for alternative auth-schemes in this case the `filter` should emit the compltete value of the Authorization header. For example : 

```json
{"token": "<some token>"}
```

```cson
filter : '"AnaplanAuthToken \(.token)"'
```

will result in a header that looks like this : 

```
Authorization: AnaplanAuthToken <some token>
```

This change is backward compatible (keep the same behaviour, make it a default that can be extended).

[1] https://datatracker.ietf.org/doc/html/rfc6750#section-2.1
[2] https://stackoverflow.com/a/53228252

## Checklist

### DIFF
--- a/tests/test_auth.py
+++ b/tests/test_auth.py
@@ -28,6 +28,38 @@ def test_custom_token_server():
 
 
 @responses.activate
+def test_custom_token_server_custom_auth_scheme():
+    """
+    RFC 2617 standard on HTTP Authentication states that credentials should
+    consist of at least a scheme and arbitrary parameters:
+        credentials = auth-scheme #auth-param
+
+    In practice auth-scheme is often "Bearer" because it is used in the OAuth2 standard.
+    This is our default. With the custom_token_server the initial request's response will be
+    injected in the the Authorization header after the Bearer auth-scheme.
+
+    This test show that we support altenative auth-schemes, in this case the `filter`
+    should emit the complete value of the Authorization header.
+    """
+    auth = Auth(
+        type='custom_token_server',
+        args=['POST', 'https://example.com'],
+        kwargs={
+            'data': {'user': 'u', 'password ': 'p'},
+            'filter': '"AnaplanAuthToken \(.token)"',  # noqa: W605
+        },
+    )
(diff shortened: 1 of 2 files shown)


tests/test_auth.py
```diff
@@ -27,6 +27,38 @@ def test_custom_token_server():
...
+    RFC 2617 standard on HTTP Authentication states that credentials should
+    consist of at least a scheme and arbitrary parameters:
+        credentials = auth-scheme #auth-param
+
+    In practice auth-scheme is often "Bearer" because it is used in the OAuth2 standard.
+    This is our default, the initial request made to the custom_token_server will be
+    used in the the Authorization header after the Bearer auth-scheme.
+
+    But we support altenative auth-schemes in this case the `filter` should emit the
+    compltete value of the Authorization header.
```
davinov, line 42: Thanks for this docstring :100:

davinov: typo: compltete > complete

davinov (approved):

tests/test_auth.py
piotch:
```suggestion
    RFC 2617 standard on HTTP Authentication states that credentials should
    consist of at least a scheme and arbitrary parameters:
        credentials = auth-scheme #auth-param

    In practice auth-scheme is often "Bearer" because it is used in the OAuth2 standard.
    This is our default, the initial request made to the custom_token_server will be
    used in the the Authorization header after the Bearer auth-scheme.

    But we support altenative auth-schemes in this case the `filter` should emit the
    complete value of the Authorization header.
```

PR #513, opened 2021-11-22
fix(mongo connector): mongo `get_slice` method now return total count in `total_rows` and returned count in `total_returned_rows`

luc-leonard:
## Change Summary

The mongo connector was not filling metadatas properly.
`total_returned_rows` was the size of the collection.
`total_rows` was empty

Now, `total_rows` is the len of the collection
`total_returned_rows` is not modified by this PR, but should be deprecated at some point

## Checklist

### DIFF
--- a/toucan_connectors/mongo/mongo_connector.py
+++ b/toucan_connectors/mongo/mongo_connector.py
@@ -238,7 +238,7 @@ class MongoConnector(ToucanConnector):
         else:
             df = self.get_df(data_source, permissions)
             total_count = len(df)
-        return DataSlice(df, stats=DataStats(total_returned_rows=total_count))
+        return DataSlice(df, stats=DataStats(total_returned_rows=len(df), total_rows=total_count))
 
     def get_df_with_regex(
         self,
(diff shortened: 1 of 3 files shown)


toucan_connectors/mongo/mongo_connector.py
```diff
@@ -238,7 +238,7 @@ def get_slice(
         else:
             df = self.get_df(data_source, permissions)
             total_count = len(df)
-        return DataSlice(df, stats=DataStats(total_returned_rows=total_count))
+        return DataSlice(df, stats=DataStats(total_returned_rows=len(df), total_rows=total_count))
```
raphaelvignes, line 241: hmmm I think to carefully test that before merging.

raphaelvignes (commented):
I guess we need to do some regression testing before merging this fix. 
(We could deploy this modification on mongo_connector on an a instance & check that everything is still working ?)

davinov (approved):
Seems great and clearer to me!
I would even remove the total_returned_rows, as I think a client will always be capable of counting the rows returned in the `data` property.

PR #519, opened 2021-11-25
fix(toucan connector): ToucanConnector#get_slice  now returns a DataSlice with `total_rows TCTC-1131

luc-leonard:


raphaelvignes (approved):
Thanks !

PR #522, opened 2021-12-07
feat: salesforce sandbox connector

raphaelvignes:
## Change Summary

Unfortunately, to use the Salesforce Service Cloud SandBox env, we need different Authorization/Token URLs. 
In oauth2 flow's current state, we cannot configure these URLs, thus they are hardcoded in Connectors Implementation. 
To enable Salesforce Sandbox usage, we have to create a new `salesforce_sandbox` connector.

## Checklist


setup.py
```diff
@@ -71,7 +71,7 @@ def get_static_file_paths():
 
 setup(
     name='toucan_connectors',
-    version='1.3.29',
+    version='1.3.30',
```
Sanix-Darker, line 74:
```suggestion
    version='1.3.31',
```
Because of this [PR](https://github.com/ToucanToco/toucan-connectors/pull/524)

toucan_connectors/salesforce_sandbox/salesforce_sandbox_connector.py
```diff
@@ -0,0 +1,194 @@
...
+        except OAuthError as ex:
+            return ConnectorStatus(status=False, error=f'Error to get status - {ex.error}')
+        except NoOAuth2RefreshToken:
+            return ConnectorStatus(
+                status=False, error='Error to get status - no refresh token found'
+            )
+        except Exception as ex:
+            return ConnectorStatus(
+                status=False, error=f'Error to get status - unknown exception - {ex}'
+            )
```
Sanix-Darker, line 194: Maybe we need also some log here ?

setup.py
raphaelvignes: yup

piotch (approved):
Approved but we discussed that this duplicates the initial connector code and we would like to change this soonish (famous last words)

PR #527, opened 2021-12-13
feat(postgres): add connector.get_status()

CharlesRngrd:
Hi Toucan,

Here is my PR to check the connection for Postgres.
I used the same logic than MySQL Connector.

**NOTE :**
I tryed to catch `psycopg2.errors` but `pgcode` is empty. So I dealt with the error message.
More infos here : https://www.psycopg.org/docs/errors.html#module-psycopg2.errors

### DIFF
--- a/tests/postgres/test_postgres.py
+++ b/tests/postgres/test_postgres.py
@@ -3,6 +3,7 @@ import psycopg2
 import pytest
 from pydantic import ValidationError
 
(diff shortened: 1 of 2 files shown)


raphaelvignes (approved):
Hello @CharlesRngrd Thanks for this PR ! 
This look ok to me :)

PR #530, opened 2021-12-20
feat(base connector): always set df memory size TCTC-1472

austil:
## Change Summary

The `df_memory_size` field of the DataStats class is so useful to know the size of the downloaded/extracted dataframe before "pagination truncation" that we should have it set all the time !

## Checklist

### DIFF
--- a/toucan_connectors/toucan_connector.py
+++ b/toucan_connectors/toucan_connector.py
@@ -362,24 +362,15 @@ class ToucanConnector(BaseModel, metaclass=ABCMeta):
             rows from a request, before limit (Snowflake)
         """
         df = self.get_df(data_source, permissions)
-        if limit is not None:
-            return DataSlice(
-                df[offset : offset + limit],
-                stats=DataStats(
-                    total_returned_rows=len(df),
-                    total_rows=len(df),
-                    df_memory_size=df.memory_usage().sum(),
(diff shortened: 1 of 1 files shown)


toucan_connectors/toucan_connector.py
```diff
@@ -362,15 +362,15 @@ def get_slice(
...
-                df[offset : offset + limit],
-                stats=DataStats(total_returned_rows=len(df), total_rows=len(df)),
-            )
-        else:
-            return DataSlice(
-                df[offset:], stats=DataStats(total_returned_rows=len(df), total_rows=len(df))
-            )
+        truncated_df = df[offset : offset + limit] if limit is not None else df[offset:]
+        return DataSlice(
+            truncated_df,
```
Sanix-Darker, line 367:
```suggestion
        return DataSlice(
            df[offset : offset + limit] if limit is not None else df[offset:],
```

Sanix-Darker (approved):
Thanks for this Review !
Just add this none-block comment !

toucan_connectors/toucan_connector.py
austil: Thx for the proposal, however I find the intermediate variable useful to get what's going on :shrug:

PR #538, opened 2021-12-22
chore: enable autocommit & release 1.3.37 for redshift

raphaelvignes:
Releasing the 1337 version + set autocommit to true in redshift connector. see: https://stackoverflow.com/q/22019154 
It was causing issues because, with autocommit off, failing queries blocked further valid queries.


Sanix-Darker (approved):
LGTM !

PR #542, opened 2022-01-06
fix(postgres): not using hostname any more (host is enough) TCTC-1603

davinov:
https://github.com/ToucanToco/toucan-connectors/pull/527 introduced a status check for postgres, that appeared alays red in cas the parameter `hostname` were used.
I actually noticed that only `host` was enough, so I suggest we only keep this one.

A migration will be needed from hostname to host for existing connections.

### DIFF
--- a/toucan_connectors/postgres/postgresql_connector.py
+++ b/toucan_connectors/postgres/postgresql_connector.py
@@ -76,15 +76,8 @@ class PostgresConnector(ToucanConnector):
 
     data_source_model: PostgresDataSource
 
-    hostname: str = Field(
-        None,
-        description='Use this parameter if you have a domain name (preferred option as more dynamic). '
-        'If not, please use the "host" parameter',
-    )
     host: str = Field(
-        None,
-        description='Use this parameter if you have an IP address. '
-        'If not, please use the "hostname" parameter (preferred option as more dynamic)',
+        None, description='The listening address of your database server (IP adress or hostname)'
     )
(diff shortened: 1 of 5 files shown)


toucan_connectors/postgres/postgresql_connector.py
```diff
@@ -124,7 +117,7 @@ def _retrieve_data(self, data_source):
 
     @staticmethod
     def _get_details(index: int, status: Optional[bool]):
-        checks = ['Hostname resolved', 'Port opened', 'Host connection', 'Authenticated']
+        checks = ['Host resolved', 'Port opened', 'Host connection', 'Authenticated']
```
Sanix-Darker, line 120: is not this redundant ?
'Host resolved', and 'Host connection'

davinov: I had the same thought, but it's actually not! The first verify that the name or IP is valid (we do this by using `self.check_hostname`), and the second one is actually the connection using `psycopg`. It's a good idea to rename it though!

davinov: b3579b8 clearer? :)

Sanix-Darker: Yes, Thanks for the clarification ! 👍🏾

Sanix-Darker (approved):
Thanks for this PR, it looks good to me !

PR #443, opened 2021-06-29
feat(query_pool): add new object to store connection

f-bb-toucantoco:
## Change Summary

Implement a ConnectionManager to keep open connection
All connections are stored in an object created at the root of the connector

`connection_manager = None
if not connection_manager:
    connection_manager = ConnectionManager(
        name='snowflake',
        timeout=20,
        wait=0.2,
        time_between_clean=10,
        time_keep_alive=600
    )
`
- timeout : Recover a connection with the timeout, if the timeout is exceeded, an exception is generated
- wait: time between 2 check to recover the connection
- time_between_clean: time between 2 laps to check if a connections needs to be removed
- time_keep_alive: time to keep the connection open

To retrieve (or create) the connection: 
`        connection = connection_manager.get(
            self.identifier,
            connect_method=lambda: connect_function(database, warehouse),
            alive_method=lambda: alive_function(),
            close_method=lambda: close_function(),
        )`

- connection_method: a method in your connector to connect at the external service
- alive_method: a method to return True or False status for the connection
- close_method: a method to close the connection at the external service

For more information about the implementation, you could check snowflake_connector.py or snowflake_oauth2_connector.py

## Change summary 2

Merge get data method in new Class for snowflake_connector.py and snowflake_oauth2_connector.py
The goals for this merge is to have only one place for Business Intelligence

## Checklist

* [X] Unit tests for the changes exist
* [X] Tests pass on CI and coverage remains at 100%
* [x] Documentation reflects the changes where applicable

PR #543, opened 2022-01-11
fix(sql): get row count for queries using aggregation methods TCTC-1567

austil:
## Change Summary

Fix a feature introduced in https://github.com/ToucanToco/toucan-connectors/pull/443 that was apparently based on wrong assumption : preventing count requests when an SQL query contained an aggregation method (sum, avg, min, etc.).

The pagination is entirely based on having the total row count. So not having it completely break the end user experience in pipeline editors.

Also, it was initialized at -1 instead of None, not a "true optional" field.
Last but not least, the regex looking for aggregation methods was not accurate enough to make the difference with the "count" method and a column named "country" :sweat_smile: 

image

## Checklist

### DIFF
--- a/tests/redshift/test_redshift.py
+++ b/tests/redshift/test_redshift.py
@@ -339,7 +339,6 @@ def test_redshiftconnector_retrieve_data(
     mock_SqlQueryHelper, mock_get_connection, redshift_connector, redshift_datasource
 ):
     mock_response = Mock()
-    mock_SqlQueryHelper.count_request_needed.return_value = True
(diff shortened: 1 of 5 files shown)


austil:
I'd like to target our backend "monthly" release branch, do I need to target something else than master for this repos ? And do a special release ?

raphaelvignes (approved):
Thanks for this Fix. 
To fix the monthly branch you only have to target master on this repo and make a release. 
In backend repo you'll need to bump the version in master & monthly.

PR #548, opened 2022-01-14
Add .devcontainer definition

raphaelvignes:
Add a dev container to be used with vscode to ease the collaboration with external contributors.

## Change Summary

Installing the repo locally can be tricky, thus the devcontainer can be use to have an independant dev environment without having to deal with dependencies issues.


raphaelvignes:
I need to add a commit so I continue here: https://github.com/ToucanToco/toucan-connectors/pull/549

PR #551, opened 2022-01-19
chore: bump from 1.3.43 to 2.0.0

Sanix-Darker:
## What
- Bump on toucan-connectors from 1.3.43 to 2.0.0.
- Added a changelog file for this repo.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Changelog (Pypi package)
 
-## [1.3.44] 2022-01-19
+## [2.0.0] 2022-01-19
 
 ### Changed
 
@@ -19,5 +19,5 @@ Some DataStats properties changed in the naming and some of them was added, see 
 - Added a dev container for developping safely on connectors.
 
 
-[1.3.44]: https://github.com/ToucanToco/toucan-connectors/compare/v1.3.40...v1.3.44
+[2.0.0]: https://github.com/ToucanToco/toucan-connectors/compare/v1.3.40...v2.0.0
 [1.3.43]: https://github.com/ToucanToco/toucan-connectors/compare/v1.3.40...v1.3.44
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,11 +2,6 @@
 
 ## [1.3.44] 2022-01-19
 
-### Added
-
-- Add filenames_to_match param to extract multiple files on connectors sharepoint and onedrive.
-- Added a dev container for developping safely on connectors.
-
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -0,0 +1,20 @@
+# Changelog (Pypi package)
+
+## [1.3.44] 2022-01-19
+
+### Added
+
```
raphaelvignes, line 6: Both of these additions were on 1.3.43
I think you can add a [1.3.43] 2022-01-18 section perhaps ? Thanks !

raphaelvignes (changes requested):
Thanks ! 
I see a small change needed in 1.3.43

CHANGELOG.md
Sanix-Darker: updated ! 👍🏾

CHANGELOG.md
```diff
@@ -0,0 +1,23 @@
...
+- Fixes on sql/snowflake (don't run count for DESCRIBE or SHOW queries + don't use -1 as default rows count)
+- Fixes on sharepoint and onedrive connectors.
+
+## [1.3.43] 2022-01-17
+
+### Added
+
+- Added filenames_to_match param to extract multiple files on connectors sharepoint and onedrive.
+- Added a dev container for developping safely on connectors.
```
davinov, line 19: It's nice, but I'm not sure we're gonna be good devs and maintain this changelog properly with every new feature added/changed or bug fixed.
I think it's more reasonable to commit on having one entry for each major version, indicating what to do regarding the breaking changes.
What do you think ? (Maybe I'm a bit too pessimistic there ahah)

CHANGELOG.md
```diff
@@ -0,0 +1,23 @@
+# Changelog (Pypi package)
+
+## [1.3.44] 2022-01-19
+
+### Changed
+
+Some DataStats properties changed in the naming and some of them was added, see [HERE](https://github.com/ToucanToco/toucan-connectors/commit/9d74efb6a6e37c6fbcd951743ac418aa84911704) for more informations.
```
davinov, line 7: Is this breaking? If yes, then we should go for a new major!

davinov (approved):
Good for me, apart from the note on the changelog scope (but I shouldn't not be the only one to decide!)

CHANGELOG.md
Sanix-Darker: 😆 i think we can give a try a least !

raphaelvignes (approved):

PR #553, opened 2022-01-26
feat(google-sheets): put magic in a new google sheet connector TCTC-1670

davinov:
## Change Summary

Repurpose the old `GoogleSheets` connector (which is not used anymore).

In this new connector:
- we don't handle the OAuth2 logic ; instead we expect an external system will provide a valid token (like Pizzly or any other OAuth2 manager)
- we use the most detailed API endpoint, so we have every formatting info about each cell ; so we can parse number and dates automagically, without the need to have them "declared" by the consumer

## Checklist


toucan_connectors/google_sheets/google_sheets_connector.py
```diff
@@ -21,51 +28,133 @@ class GoogleSheetsDataSource(ToucanDataSource):
...
+    It needs to be provided a retrieve_token method which should provide a valid OAuth2 access token.
+    Not to be confused with the OAuth2 connector, which handles all the OAuth2 process byt itself!
     """
 
     data_source_model: GoogleSheetsDataSource
-    bearer_integration = 'google_sheets'
-    bearer_auth_id: str
+
+    _auth_flow = 'managed_oauth2'
+    _retrieve_token: Callable[[str], str] = PrivateAttr()
```
raphaelvignes, line 43: Smart, this method will be delegated from `laputa` ?

doc/connectors/google_sheets.md
```diff
@@ -4,20 +4,17 @@
 
 * `type`: `"GoogleSheets"`
 * `name`: str, required
-* `bearer_auth_id`: str, required
+* `auth_id`: str, required
 
-The `bearer_auth_id` is a token retrieved by the front-end
-to let Bearer handle the OAuth dance for you.
+Note: this connector needs a `retrieve_token` function that will be given the `auth_id` and should return a valid token.
```
raphaelvignes, line 9: From where is coming this function ?

raphaelvignes (commented):
Thanks for this refurbishment!
Needs some lint! I've got few questions (sorry if the answer is obvious, I'm might not be up to date on this topic :s )

doc/connectors/google_sheets.md
davinov: This will be provided by the host (Laputa in our case)

davinov:
I've reworked a bit so it uses a more consistent API (the detailed API may have empty rows, cells - the values API always return the expected array)

raphaelvignes (approved):
Thanks for this PR. 
I see a couple of #TODO & #FIXME, I guess we should provision some cards in the backlog to handle it.

PR #554, opened 2022-01-28
fix(sql): pandas read_sql needs SQLAlchemy

davinov:
Tests were failing because this dependency is not installed: https://github.com/ToucanToco/toucan-connectors/runs/4969849017?check_suite_focus=true


raphaelvignes (approved):
Thx :+1:

PR #566, opened 2022-02-09
fix(cache key) found the real culprit for the wrong cache key: permissions had to be resolved TCTC-1882

luc-leonard:


toucan_connectors/toucan_connector.py
```diff
@@ -451,15 +448,16 @@ def get_cache_key(
         This identifier will then be used as a cache key.
         """
         unique_identifier = {
-            'connector': self.get_unique_identifier(data_source),
-            'permissions': permissions,
+            'connector': self.get_unique_identifier(),
+            'permissions': nosql_apply_parameters_to_query(permissions, data_source.parameters)
```
raphaelvignes, line 452: :+1:

raphaelvignes (approved):
Thx !

PR #569, opened 2022-02-15
fix(postgres): format all result from get model

raphaelvignes:
## Change Summary
issue on postgres: only the last DB tree was returned when calling `get-model`


toucan_connectors/postgres/postgresql_connector.py
```diff
@@ -182,4 +182,4 @@ def get_model(self):
                     cursor.execute(build_database_model_extraction_query())
                     res = cursor.fetchall()
                     databases_tree += res
```
Sanix-Darker, line 184:
```suggestion
                    databases_tree += cursor.fetchall()
```

Sanix-Darker (approved):

PR #574, opened 2022-02-18
fix(snowflake): get model query

raphaelvignes:
Fix the issue on duplicate columns displayed in Snowflake's Graphical interface.

### DIFF
--- a/toucan_connectors/snowflake_common.py
+++ b/toucan_connectors/snowflake_common.py
@@ -65,7 +65,7 @@ def build_database_model_extraction_query() -> str:
     from
         information_schema.tables t
     inner join information_schema.columns c on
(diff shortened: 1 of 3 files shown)


toucan_connectors/snowflake_common.py
```diff
@@ -65,7 +65,7 @@ def build_database_model_extraction_query() -> str:
     from
         information_schema.tables t
     inner join information_schema.columns c on
-        t.table_name = c.table_name
+        t.table_name = c.table_name and t.table_schema = c.table_schema
```
raphaelvignes, line 68: We can have tables with the same names in != schemas

jGundermann (approved):
LGTM. 
If I were to be picky, I would add a line break.

PR #575, opened 2022-02-22
fix: support custom variables syntax in default cache key TCTC-1903

austil:
## Change Summary

Fix a issue on Google Big Queries request using the custom variable syntax `@my_var` 
by making the default cache key "dumber" to produce different cache key for queries using custom variable syntax
(other than `{{ }}` & `@()`.

Resolution path considered but not chosen includes : 

- dropping support for this custom syntax on user side -> would require checking is the query does not use this syntax & do this for every connector with specific variable syntax, nope
-  overwriting `_render_datasource` for BigQuery like it was done for Mongo, Snowflake & Http -> apparently not possible, the BigQuery client does not seem to have a method returning interpolated query.

## Checklist

### DIFF
--- a/toucan_connectors/toucan_connector.py
+++ b/toucan_connectors/toucan_connector.py
@@ -428,9 +428,9 @@ class ToucanConnector(BaseModel, metaclass=ABCMeta):
 
     def _get_unique_datasource_identifier(self, data_source: ToucanDataSource) -> dict:
         # By default we don't know which variable syntax is be supported by the inheriting connector,
-        # so calling `nosql_apply_parameters_to_query` is wrong and will produce the same cache key 
+        # so calling `nosql_apply_parameters_to_query` is wrong and will produce the same cache key
         # for different queries when using custom variable syntax !
-        # Overwrite this method to improve the cache key at places where supported syntaxes are clear. 
+        # Overwrite this method to improve the cache key at places where supported syntaxes are clear.
         return data_source.dict()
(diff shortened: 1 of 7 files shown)


toucan_connectors/toucan_connector.py
```diff
@@ -426,13 +426,12 @@ def get_unique_identifier(self) -> str:
...
-            data_source.dict(), data_source.parameters, handle_errors=True
-        )
-        del data_source_rendered['parameters']
-
-        return data_source_rendered
+    def _get_unique_datasource_identifier(self, data_source: ToucanDataSource) -> dict:
+        # By default we don't know which variable syntax is be supported by the inheriting connector,
+        # so calling `nosql_apply_parameters_to_query` is wrong and will produce the same cache key
+        # for different queries when using custom variable syntax !
+        # Overwrite this method to improve the cache key at places where supported syntaxes are clear.
```
raphaelvignes, line 433: :fire: thanks for this docstring !

raphaelvignes (approved):
So `_render_datasource` becomes useless :rofl:  
Thanks for this Fix/Update :+1:

toucan_connectors/toucan_connector.py
davinov: Bets to use `"""` for doctrings :) (IDEs and doc generators can use doc strings defined with triple quotes, but not single-line comments)

PR #584, opened 2022-04-19
feat: databricks connector handle not started cluster

raphaelvignes:
* Dedicated Databricks connector
* Handle case when the cluster is not started.


toucan_connectors/databricks/databricks_connector.py
```diff
@@ -0,0 +1,129 @@
...
+                raise DataBricksUnauthorizedError('Invalid credentials')
+            else:
+                raise DataBricksConnectionError('Invalid connection params')
+
+    def _build_connection_string(self) -> str:
+        """For constants see: https://docs.databricks.com/integrations/bi/jdbc-odbc-bi.html"""
+        connection_params = {
+            'Driver': '/opt/simba/spark/lib/64/libsparkodbc_sb64.so',  # This path must be correct when installing
+            # ODBC driver
+            'Host': '127.0.0.0.1' if self.Host == 'localhost' else self.Host,
```
Sanix-Darker, line 69:
```suggestion
            'Host': '127.0.0.1' if self.Host == 'localhost' else self.Host,
```

Sanix-Darker: And maybe handle the case of `0.0.0.0` ?

toucan_connectors/databricks/databricks_connector.py
```diff
@@ -0,0 +1,129 @@
...
+            self._connect()
+        except DataBricksConnectionError as e:
+            return ConnectorStatus(
+                status=False, details=self._get_details(2, False), error=e.args[0]
+            )
+
+        except (DataBricksUnauthorizedError, Exception) as e:
+            return ConnectorStatus(
+                status=False, details=self._get_details(3, False), error=e.args[0]
+            )
```
Sanix-Darker, line 110: IMO these Exceptions catch should be directly inside the  `_connect` method, what do you think ?

raphaelvignes: No, I don't think so, we specifically raise them in `_connect` to provide the correct details in `get_status`

toucan_connectors/databricks/databricks_connector.py
raphaelvignes: thx, I guess we can rewrite it to '127.0.0.1'

raphaelvignes:
unfortunately we don't have `offset` keyword in Databricks as it is SparKsql dialect.

toucan_connectors/databricks/databricks_connector.py
```diff
@@ -0,0 +1,128 @@
...
+    Port: int = Field(
+        ..., description='The listening port of your databricks cluster', placeholder=443
+    )
+    HTTPPath: str = Field(
+        ..., description='Databricks compute resources URL', placeholder='sql/protocolv1/o/xxx/yyy'
+    )
+    PWD: SecretStr = Field(
+        None, description='Your personal access token', placeholder='dapixxxxxxxxxxx'
+    )
+    Ansi: bool = False
```
lukapeschke, line 46: Est-ce qu'on veut vraiment du non-snake-case pour les champs du connecteur ?

lukapeschke (commented):
apart from the nit on DatabricksConnector fields, LGTM

toucan_connectors/databricks/databricks_connector.py
```diff
@@ -0,0 +1,131 @@
...
+            'ThriftTransport': 2,
+            'SSL': 1,
+            'AuthMech': 3,
+            'UID': 'token',
+            'PWD': self.pwd.get_secret_value(),
+        }
+        return ';'.join(f'{k}={v}' for k, v in connection_params.items() if v is not None)
+
+    @staticmethod
+    def _get_details(index: int, status: Optional[bool]):
```
lukapeschke, line 69: Could you add typing please ?

Also, maybe the method would be mode explicit in the code if it was called with the check's name rather than its index in a list ?

toucan_connectors/databricks/databricks_connector.py
```diff
@@ -0,0 +1,131 @@
...
+        # Check port
+        try:
+            self.check_port(self.host, self.port)
+        except Exception as e:
+            return ConnectorStatus(status=False, details=self._get_details(1, False), error=str(e))
+        try:
+            pyodbc.connect(self._build_connection_string())
+        except pyodbc.InterfaceError as e:
+            details = self._get_details(3, False)
+            details[2] = ('Connected to Databricks', True)
```
lukapeschke, line 91: Is this really needed ? Shouldn't it already be set to `True` if `index` is 3 ?

toucan_connectors/databricks/databricks_connector.py
```diff
@@ -0,0 +1,131 @@
...
+
+    def start_cluster(self) -> None:
+        endpoint = f'https://{self.host}/api/2.0/clusters/start'
+        headers = {'login': 'token', 'password': self.pwd.get_secret_value()}
+        data = {'cluster_id': self.http_path.split('/')[-1]}
+        resp = requests.post(endpoint, headers=headers, json=data)
+        if resp.status_code == 200:
+            logger.info('Databricks cluster started')
+        else:
+            logger.info(resp.json().get('message', 'Failed to start Databricks cluster'))
```
lukapeschke, line 113: It could be nice for the caller to have the information that the cluster did not start. Maybe raise an exception here ?

tests/databricks/test_databricks.py
```diff
@@ -0,0 +1,213 @@
...
+        method='POST',
+        url='https://127.0.0.1/api/2.0/clusters/start',
+        headers={'login': 'token', 'password': '12345'},
+        match=[responses.matchers.json_params_matcher({'cluster_id': 'path'})],
+        json={},
+    )
+    sypied_start = mocker.spy(DatabricksConnector, 'start_cluster')
+    databricks_connector.start_cluster()
+    sypied_start.assert_called_once()
+
```
lukapeschke, line 195: I don't get the point of this ? We call the method and check that it has been called ?

lukapeschke: Or is it to ensure that it has not been called twice ?

tests/databricks/test_databricks.py
```diff
@@ -0,0 +1,213 @@
...
+        url='https://127.0.0.1/api/2.0/clusters/start',
+        headers={'login': 'token', 'password': '12345'},
+        match=[responses.matchers.json_params_matcher({'cluster_id': 'path'})],
+        json={'message': 'Failed to start Databricks cluster'},
+        status=400,
+    )
+    sypied_start = mocker.spy(DatabricksConnector, 'start_cluster')
+    mockedlog = mocker.patch('toucan_connectors.databricks.databricks_connector.logger.info')
+    databricks_connector.start_cluster()
+    sypied_start.assert_called_once()
```
lukapeschke, line 212: Same here, I'm not sure to understand the point of this

lukapeschke (commented):
A few remarks here and there but implementation LGTM :+1:

toucan_connectors/databricks/databricks_connector.py
raphaelvignes: :+1:

toucan_connectors/databricks/databricks_connector.py
raphaelvignes: yep, forgot to clean that :+1:

toucan_connectors/databricks/databricks_connector.py
raphaelvignes: yes, I add that.

tests/databricks/test_databricks.py
raphaelvignes: nope just a mistake. Ensuring that the correct call to URL is made is enough

lukapeschke (approved):
Thanks for the changes, LGTM! :+1:

raphaelvignes:
Thx, j'attends de valider la PR laputa avant de merge

toucan_connectors/databricks/databricks_connector.py
```diff
@@ -0,0 +1,136 @@
+import logging
+from typing import Dict, List, Optional, Tuple
```
lukapeschke, line 2: `Dict` is unused, makes the linter fail

PR #588, opened 2022-04-25
chore: v3.3.1

raphaelvignes:
Bump connectors v3.3.1:
- Fix on Athena Connector
- Databricks Connector

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -14,7 +14,6 @@ extras_require = {
     'azure_mssql': ['pyodbc>=3'],
     'clickhouse': ['clickhouse_driver', 'sqlalchemy'],
     'dataiku': ['dataiku-api-client'],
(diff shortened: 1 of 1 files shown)


Sanix-Darker (changes requested):
Thanks for the bump, you forgot to update sonar-project.properties file

Sanix-Darker (approved):

lukapeschke:
`get_status` has been added to the athena connector since 3.3.0, so shouldn't this be 3.4.0 ?

Sanix-Darker:
good point @lukapeschke with the new databrick connector.... it should be a 'major' version upgrade, what do you think @raphaelvignes ?

lukapeschke (approved):

PR #589, opened 2022-04-25
fix: databricks get status

raphaelvignes:
## Change Summary
Databricks `get_status` needs to use an autocommit connection.


lukapeschke (approved):
:+1:

PR #590, opened 2022-05-04
feat(googlebigquery): parse dates as datetime dtype [TCTC-2521]

fspot:
## Change Summary

without date_as_object=False, date columns get dtype 'object' instead of 'datetime64[ns]'


PrettyWood (changes requested):
can we have a unittest?

davinov (approved):
Not sure it deserve a unit test (it's just calling a function with a parameter, not really some business logic).
But it would be nice to have some integration tests for these, to help us confidently see in advance:
- the impact of upgrading libraries
- if the service we target has changed its API in some way

In short: not a topic for this PR but for a general issue for the repo

PrettyWood (approved):

PR #591, opened 2022-05-09
fix(sql connectors): set `datetime` dtype when needed [TCTC-1576] [TCTC-2521]

fspot:
## Change Summary

Adds an util `infer_datetime_dtype` and use it for each sql connector. Its purpose is to properly set the dtype on datetime columns.

Performance impact : 4ms on *shark attack* dataset (6750 rows, 16 columns).


jGundermann:
LGTM

jGundermann (approved):

PR #594, opened 2022-05-11
feat(google BigQuery): Add support for array query parameters TCTC-1447

lukapeschke:
## Change Summary

This allows to pass array query parameters. These were defined as strings until now,
which caused issues when using them with BigQuery functions expecting an array, such
as UNNEST.

## Checklist

### DIFF
--- a/toucan_connectors/google_big_query/google_big_query_connector.py
+++ b/toucan_connectors/google_big_query/google_big_query_connector.py
@@ -1,7 +1,7 @@
 import logging
 from enum import Enum
 from timeit import default_timer as timer
-from typing import Dict, List, Optional
+from typing import Any, Dict, List, Optional, Union
 
 import pandas
 import pandas as pd
@@ -27,8 +27,14 @@ class GoogleBigQueryDataSource(ToucanDataSource):
     )
 
 
-# TODO - check bad return type
-def _define_type(value) -> str:
+BigQueryParam = Union[bigquery.ScalarQueryParameter, bigquery.ArrayQueryParameter]
+
+
+# NOTE: This does not play nicely with dates. They're a bit tricky
+# though, as we'd have to try and parse dates from strings to
+# determine if something is a date or not. Until then, we can just
+# use a cast. eg: SELECT * FROM table WHERE STRING(date_col) IN UNNEST({{my_dates}})
+def _define_scalar_type(value: Any) -> str:
     if isinstance(value, bool):
         return 'BOOL'
     elif isinstance(value, int):
@@ -37,8 +43,21 @@ def _define_type(value) -> str:
         return 'FLOAT64'
     elif isinstance(value, str):
         return 'STRING'
+    # TODO - check bad return type
+    return 'STRING'
+
+
+def _define_array_type(name: str, values: List[Any]) -> BigQueryParam:
+    return bigquery.ArrayQueryParameter(
+        name, _define_scalar_type(values[0] if len(values) > 0 else ''), values
+    )
+
+
+def _define_query_param(name: str, value: Any) -> BigQueryParam:
+    if isinstance(value, list):
+        return _define_array_type(name, value)
     else:
-        return 'STRING'
+        return bigquery.ScalarQueryParameter(name, _define_scalar_type(value), value)
 
 
 class GoogleBigQueryConnector(ToucanConnector):
@@ -119,13 +138,14 @@ class GoogleBigQueryConnector(ToucanConnector):
 
     @staticmethod
     def _prepare_parameters(query: str, parameters: Optional[Dict]) -> List:
+        """replace ToucanToco variable definitions by Google Big Query variable definition"""
         query_parameters = []
-        """replace ToucanToco variable definition by Google Big Query variable definition"""
-        for k in parameters or {}:
-            if query.find('@' + k) > -1:
+        for param_name, param_value in (parameters or {}).items():
+            if query.find('@' + param_name) > -1:
                 # set all parameters with a type defined and necessary for Big Query
                 query_parameters.append(
-                    bigquery.ScalarQueryParameter(k, _define_type(parameters[k]), parameters[k])
+                    _define_query_param(param_name, param_value)
+                    # bigquery.ScalarQueryParameter(k, _define_type(parameters[k]), parameters[k])
                 )
         return query_parameters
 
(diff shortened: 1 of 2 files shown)


toucan_connectors/google_big_query/google_big_query_connector.py
```diff
@@ -119,13 +138,14 @@ def _execute_query(client: bigquery.Client, query: str, parameters: List) -> pan
...
-        """replace ToucanToco variable definition by Google Big Query variable definition"""
-        for k in parameters or {}:
-            if query.find('@' + k) > -1:
+        for param_name, param_value in (parameters or {}).items():
+            if query.find('@' + param_name) > -1:
                 # set all parameters with a type defined and necessary for Big Query
                 query_parameters.append(
-                    bigquery.ScalarQueryParameter(k, _define_type(parameters[k]), parameters[k])
+                    _define_query_param(param_name, param_value)
+                    # bigquery.ScalarQueryParameter(k, _define_type(parameters[k]), parameters[k])
```
davinov, line 148: forgot to remove that comment?

toucan_connectors/google_big_query/google_big_query_connector.py
```diff
@@ -37,8 +43,21 @@ def _define_type(value) -> str:
         return 'FLOAT64'
     elif isinstance(value, str):
         return 'STRING'
+    # TODO - check bad return type
+    return 'STRING'
+
+
+def _define_array_type(name: str, values: List[Any]) -> BigQueryParam:
+    return bigquery.ArrayQueryParameter(
+        name, _define_scalar_type(values[0] if len(values) > 0 else ''), values
```
davinov, line 52: Obviously this will only work for arrays that have all elements of the same type.
It should probably be the case for the vast majority though, so it seems OK.

toucan_connectors/google_big_query/google_big_query_connector.py
raphaelvignes, line 148: To clean?

toucan_connectors/google_big_query/google_big_query_connector.py
```diff
@@ -27,8 +27,14 @@ class GoogleBigQueryDataSource(ToucanDataSource):
...
 
-# TODO - check bad return type
-def _define_type(value) -> str:
+BigQueryParam = Union[bigquery.ScalarQueryParameter, bigquery.ArrayQueryParameter]
+
+
+# NOTE: This does not play nicely with dates. They're a bit tricky
+# though, as we'd have to try and parse dates from strings to
+# determine if something is a date or not. Until then, we can just
+# use a cast. eg: SELECT * FROM table WHERE STRING(date_col) IN UNNEST({{my_dates}})
```
raphaelvignes, line 36: Perhaps, this is a larger topic about how to handle dates in data sources parameters. 
Your solution seems fine for now.

raphaelvignes (approved):
Thx for this fix. 
I see a comment that could be cleaned I guess.

toucan_connectors/google_big_query/google_big_query_connector.py
lukapeschke: oops yes, good catch!

davinov (approved):
Thanks! The provided change is OK :+1: 

Still, reading the connector's code and how we handle the parameters raises a few questions that I'm adding as comment outside this review (below)

toucan_connectors/google_big_query/google_big_query_connector.py
```diff
@@ -27,8 +27,14 @@ class GoogleBigQueryDataSource(ToucanDataSource):
...
+def _define_scalar_type(value: Any) -> str:
```
davinov, line 37: I dug a bit into the bigquery lib and found https://github.com/googleapis/python-bigquery/blob/v2.34.3/google/cloud/bigquery/dbapi/_helpers.py#L68 which seem to do the same job.
I went to see that because I was intrigued when I read [this doc](https://googleapis.dev/python/bigquery/2.34.3/dbapi.html#providing-explicit-type-information) saying that using the DB-API, the lib would determine the parameters type itself.

Maybe that's worth a shot to simplify our connector?

toucan_connectors/google_big_query/google_big_query_connector.py
davinov, line 36: Thanks for this comment :+1: 
However, it does seem to have a DATE (and even a TIMESTAMP) type available: https://googleapis.dev/python/bigquery/latest/query.html#google.cloud.bigquery.query.ScalarQueryParameterType
Can't we use them?

lukapeschke: Sure we could use them, but to me the problem lies with the format in which we receive the query parameters from the frontend: since we receive them as JSON, dates will be strings. Parameters are provided as a `{param_name: value}` dict, without any metadata about their type. How do we determine if a string is a date or a regular string ? Should we try to parse a date from a pre-defined list of date formats and if none of them works, consider the param to be a string ? This could be tricky, as we might have false positives, like a `"202001"` string representing an ID being interpreted as january 2020

toucan_connectors/google_big_query/google_big_query_connector.py
lukapeschke: Yeah that's flaky. Worst thing is that it probably won't fail if we have an array of mixed types, since everything gets serialized as JSON again. So everything will probably be cast to a string or something like that.

And the bigquery lib seems to do the same as us here :stuck_out_tongue: https://github.com/googleapis/python-bigquery/blob/35aeaa61db271027cb95d68de362544e437dd761/google/cloud/bigquery/dbapi/_helpers.py#L119

toucan_connectors/google_big_query/google_big_query_connector.py
lukapeschke: Nice, thanks :+1: Is it fine with you if I merge this as is, and make another refactor PR to use these functions rather than ours ?

toucan_connectors/google_big_query/google_big_query_connector.py
davinov: Good point!
We should be able to receive parameters as "real dates": we added a layer, in our HTTP API, that serializes and de-serializes dates as `{$date: <timestamp>}`. (Inspired by Mongo's extended JSON specification)

lukapeschke: Aaaah nice! So it should be "auto-handled" by #595

PR #601, opened 2022-05-19
chore(deps): Pin googleads>11.0.0

lukapeschke:
This release removed the suds-jurko dependency, which is unmaintained, and
a pain to install.

https://github.com/googleads/googleads-python-lib/

FYI, I checked on our deployment, and since the versions were unspecified anyway, we have `googleads` 31.0.0 installed (11.0 is from 2018)


raphaelvignes (approved):
thx

PR #603, opened 2022-05-24
feat(mongo-connector): allow search on multiple columns [TCTC-2532]

aklom:
## Description 
With the `get_df_with_regex` we could only search for a string on a single field given as a param. We couldn't also get the whole slice of data. That's why we needed to create a new function that returns the slice and allows searching for a string on multiple columns given as an array of strings in the params. 

Feature needed for https://toucantoco.atlassian.net/jira/software/c/projects/TCTC/boards/13?modal=detail&selectedIssue=TCTC-2532&quickFilter=87

### DIFF
--- a/toucan_connectors/mongo/mongo_connector.py
+++ b/toucan_connectors/mongo/mongo_connector.py
@@ -250,7 +250,7 @@ class MongoConnector(ToucanConnector):
         permissions: Optional[str] = None,
         limit: Optional[int] = None,
         offset: Optional[int] = None,
-    ) -> pd.DataFrame:
+    ) -> DataSlice:
         # Create a copy in order to keep the original (deepcopy-like)
         data_source = MongoDataSource.parse_obj(data_source)
         data_source.query = normalize_query(data_source.query, data_source.parameters)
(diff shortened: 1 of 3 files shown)


toucan_connectors/mongo/mongo_connector.py
```diff
@@ -242,13 +242,14 @@ def get_slice(
...
+    def get_slice_with_regex(
         self,
         data_source: MongoDataSource,
-        field: str,
+        fields: List[str],
         regex: Pattern,
         permissions: Optional[str] = None,
         limit: Optional[int] = None,
+        offset: Optional[int] = None,
     ) -> pd.DataFrame:
```
lukapeschke, line 253:
```suggestion
    ) -> DataSlice:
```

You're not returning the `.df` attribute anymore

lukapeschke (approved):
:+1:

toucan_connectors/mongo/mongo_connector.py
```diff
@@ -259,15 +260,38 @@ def get_df_with_regex(
...
             }
         }
         data_source.query.append({'$match': regex_step})
-        return self.get_slice(data_source, permissions, limit=limit).df
+        return self.get_slice(data_source, permissions, limit=limit, offset=offset)
+
+    def get_df_with_regex(
+        self,
+        data_source: MongoDataSource,
+        field: str,
```
PrettyWood, line 281: Why not support list here too? I don't understand.

skornous: Because it wasn't supported in the original source code so we kept the function's contract

PR #606, opened 2022-06-08
chore: Move to poetry TCTC-2931

lukapeschke:


pyproject.toml
```diff
@@ -1,4 +1,170 @@
...
+version = "3.8.0"
+description = "Toucan Toco Connectors"
+authors = ["Toucan Toco <dev@toucantoco.com>"]
+license = "BSD"
+readme = "README.md"
+
+[tool.poetry.dependencies]
+# <3.11 required by awswrangler
+python = ">=3.8,<3.11"
+Authlib = "^1.0.1"
```
Sanix-Darker, line 22: IMO, we should have for all our dependencies  `>=` instead of `^`, related discussion on laputa with @fspot : https://github.com/ToucanToco/laputa/pull/5135#discussion_r879380638

Makefile
```diff
@@ -1,6 +1,7 @@
 .DEFAULT_GOAL := all
-isort = isort toucan_connectors tests setup.py
-black = black toucan_connectors tests setup.py
+isort = isort toucan_connectors tests
+black = black toucan_connectors tests
+flake8 = flake8 toucan_connectors tests
```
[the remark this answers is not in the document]
lukapeschke: We don't do it in laputa, and I find it more constraining: 
* If the user wants to manage their venvs themselves rather than delegating that to poetry (which most of us have to do to juggle between python versions), having `poetry run` is annoying.
* If the user does want to delegate venv creation to poetry, they can easily activate it before running `make` commands, either with manual activation or with `poetry shell`

pyproject.toml
lukapeschke: I disagree, especially for stable libs such as `Authlib` or `psycopg2`, were suddenly having a new major version in production images is definitely not something we want. But you're right for 0ver, I'll update everything in 0ver to be `>=0.x,<1.0`

jGundermann (approved):
LGTM

fspot (approved):
:pray:

PR #609, opened 2022-06-13
feat: add support for elasticsearch 8 [TCTC-1907]

Sanix-Darker:
## Change Summary
- Unpin + bump elasticsearch from 7.17.4 to 8.2.2
- Updated the connector and tests due to the `send_get_body_as` param deprecated on 7.16 (see [HERE](https://github.com/elastic/elasticsearch-py/releases?page=3#:~:text=Deprecated%20the%20send_get_body_as%20parameter.%20This%20parameter%20is%20no%20longer%20necessary%0Aas%20APIs%20all%20use%20non%2DGET%20HTTP%20methods%20when%20using%20a%20body.))

## Checklist

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -38,7 +38,7 @@ awswrangler = {version = "^2.15.1", optional = true}
 pyodbc = {version = "^4", optional = true}
 clickhouse-driver = {version = ">=0.2.3,<1.0", optional = true}
 dataiku-api-client = {version = "^9.0.1", optional = true}
-elasticsearch = {version = ">=7.16.0,<8", optional = true}
+elasticsearch = {version = ">=7.11.0,<8", optional = true}
 facebook-sdk = {version = "^3.1.0", optional = true}
 python-graphql-client = {version = ">=0.4.3,<1.0", optional = true}
 google-api-python-client = {version = "^2", optional = true}
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -38,7 +38,7 @@ awswrangler = {version = "^2.15.1", optional = true}
 pyodbc = {version = "^4", optional = true}
 clickhouse-driver = {version = ">=0.2.3,<1.0", optional = true}
 dataiku-api-client = {version = "^9.0.1", optional = true}
-elasticsearch = {version = ">=8", optional = true}
+elasticsearch = {version = ">=7.16.0,<8", optional = true}
 facebook-sdk = {version = "^3.1.0", optional = true}
 python-graphql-client = {version = ">=0.4.3,<1.0", optional = true}
 google-api-python-client = {version = "^2", optional = true}
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -38,7 +38,7 @@ awswrangler = {version = "^2.15.1", optional = true}
 pyodbc = {version = "^4", optional = true}
 clickhouse-driver = {version = ">=0.2.3,<1.0", optional = true}
 dataiku-api-client = {version = "^9.0.1", optional = true}
-elasticsearch = {version = "<8", optional = true}
+elasticsearch = {version = ">=8", optional = true}
 facebook-sdk = {version = "^3.1.0", optional = true}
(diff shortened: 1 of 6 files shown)


raphaelvignes (approved):
Thx for the update :+1:

lukapeschke (changes requested):
Seems dangerous to me. This is a very recent ES version, lots of deployments are still 6.x or 7.x (even though 6.x is deprecated, so >=7 seems reasonnable)

lukapeschke:
> Seems dangerous to me. This is a very recent ES version, lots of deployments are still 6.x or 7.x (even though 6.x is deprecated, so >=7 seems reasonnable)

Sorry for the short comment, just wanted to make sure we don't merge this before discussion. Maybe we could provide two extras, one for ES 7 and one for ES 8 ?

Sanix-Darker:
> lots of deployments are still 6.x or 7.x 
What do you mean by that ? 🤔 @lukapeschke 

Correct me if am wrong, but this is a connector bump, not an infra bump, this should have nothing to do with our current stack !

In this PR, we're bumping the elasticsearch lib, tests related to it (the feeding part (because REST API endpoints changed)) and the image from CI, that doesn't have impact anywhere else... 🤔 

But yeah, this may need a discussion, idk, maybe am missing something...

lukapeschke:
> > lots of deployments are still 6.x or 7.x
> 
> What do you mean by that ? thinking @lukapeschke

A lot of Elasticsearch deployments out there are still in V6 or V7. We don't care about V6 since it's EOL, but we should definitely support V7 as I'm afraid some customers will complain that toucan can't be plugged to their cluster otherwise.

> 
> Correct me if am wrong, but this is a connector bump, not an infra bump, this should have nothing to do with our current stack !

Elasticsearch bumps all their products simultaneously, so ES clusters in v7 require `elasticsearch-py >=7,<8` and clusters in v8 require `elasticsearch-py >=8,<9`. This might (or might not, depending on their config) break the connectors of all clients 
who are still running elastic in v7

> 
> In this PR, we're bumping the elasticsearch lib, tests related to it (the feeding part (because REST API endpoints changed)) and the image from CI, that doesn't have impact anywhere else... thinking

Bumping espy has an impact on everyone using the elasticsearch connector. As you said, the endpoints change between v7 and v8, so we need to make sure this still works with ES 7

Sanix-Darker:
> lots of deployments are still 6.x or 7.x

@lukapeschke , do you have an idea if there is a way to have real life statistics for this ? 🤔

pyproject.toml
```diff
@@ -38,7 +38,7 @@ awswrangler = {version = "^2.15.1", optional = true}
 pyodbc = {version = "^4", optional = true}
 clickhouse-driver = {version = ">=0.2.3,<1.0", optional = true}
 dataiku-api-client = {version = "^9.0.1", optional = true}
-elasticsearch = {version = "<8", optional = true}
+elasticsearch = {version = ">=7.16.0,<8", optional = true}
```
lukapeschke, line 41:
```suggestion
elasticsearch = {version = ">=7.11.0,<8", optional = true}
```
7.11 isn't EOL yet

tests/docker-compose.yml
```diff
@@ -1,7 +1,9 @@
 elasticsearch:
-    image: docker.elastic.co/elasticsearch/elasticsearch:7.0.0
+    image: docker.elastic.co/elasticsearch/elasticsearch:7.16.0
```
lukapeschke, line 2: I think it is important to also run tests against elasticsearch 8. Want me to add them ?

pyproject.toml
Sanix-Darker: https://github.com/ToucanToco/toucan-connectors/pull/609/commits/d9b99d683ffdd829ecc02edb00b4067bb8083a58

tests/docker-compose.yml
Sanix-Darker: Yes please, do so 🙏🏾

lukapeschke (approved):

PR #615, opened 2022-06-23
chore: v3.12.0

raphaelvignes:
Bump to v3.12.0


lukapeschke (commented):
We've lost a feature, so I'd rather have a v3.12.0. LGTM apart from that

jGundermann:
Shouldn't we also update the Changelog?

PR #621, opened 2022-06-27
fix: make exploration faster + add form

PrettyWood:


toucan_connectors/redshift/redshift_database_connector.py
```diff
@@ -347,39 +367,20 @@ def describe(self, data_source: RedshiftDataSource) -> dict[str, Any]:
...
-                ORDER BY
-                    table_schema,
-                    table_name,
-                    ordinal_position;
-                """
+    def _db_table_info_rows(self, database: str) -> list[tuple[str, str, str, str]]:
+        with self._get_cursor(database) as cursor:
+            """Get rows of (schema, table name, column name, column type)"""
+            cursor.execute(
+                r"""SELECT "schemaname", "tablename", "column", "type" FROM PG_TABLE_DEF WHERE schemaname = 'public';"""
```
raphaelvignes, line 374: Could fail if the db user is not allowed to query this table. 
But it's okay to try IMO.

toucan_connectors/redshift/redshift_database_connector.py
```diff
@@ -133,11 +148,16 @@ class RedshiftConnector(ToucanConnector, DiscoverableConnector):
 
     class Config:
         underscore_attrs_are_private = True
+        keep_untouched = (cached_property,)
 
         @staticmethod
         def schema_extra(schema: dict[str, Any]) -> None:
             schema['properties'] = {k: schema['properties'][k] for k in ORDERED_KEYS}
 
+    @cached_property
```
raphaelvignes, line 157: :+1:

raphaelvignes (approved):
Thx!

tests/redshift/test_redshift.py
```diff
@@ -472,17 +472,18 @@ def test_redshiftconnector_describe(mock_connection, redshift_connector, redshif
...
-        ['dev', 'public', 'table_1', 'value1', 'bigint'],
-        ['dev', 'public', 'table_2', 'label', 'character varying'],
-        ['dev', 'public', 'table_2', 'doum', 'character varying'],
-        ['dev', 'public', 'table_2', 'value1', 'bigint'],
-        ['dev', 'public', 'table_2', 'value2', 'bigint'],
-        ['dev', 'public', 'table_3', 'label', 'character varying'],
-        ['dev', 'public', 'table_3', 'group', 'character varying'],
+    mocker.patch.object(RedshiftConnector, '_list_db_names', return_value=['dev'])
+    mocker.patch.object(RedshiftConnector, '_db_table_info_rows').return_value = [
+        ('pg_internal', 'redshift_auto_health_check_436837', 'a', 'integer'),
```
PrettyWood, line 477: I should probably remove this line on the test. I decided to add the  `WHERE schemaname = 'public';`. Waiting for your opinion. But in term of perf it can be super slow to retrieve all the internal information and I don't expect anyone to use them

raphaelvignes: Is it possible to have another schema than public having non system information ?

raphaelvignes: Let's stick to public schema for now.

raphaelvignes:
Good for me :+1:

toucan_connectors/redshift/redshift_database_connector.py
jGundermann: one thought about it is when data get's large the cached property stay's forever in RAM.
```suggestion
    @cached_property_with_ttl(ttl=300) # five minutes ttl
```

PR #634, opened 2022-07-22
feat: catch error on mongo connection

Sanix-Darker:
## WHAT

- catch errors on awsathena when getting an attribute from NoneType
- catch ConfigurationError on mongo-connector

### DIFF
--- a/tests/docker-compose.yml
+++ b/tests/docker-compose.yml
@@ -1,3 +1,5 @@
+---
+
 elasticsearch7:
   image: docker.elastic.co/elasticsearch/elasticsearch:7.16.0
   environment:
--- a/tests/docker-compose.yml
+++ b/tests/docker-compose.yml
@@ -1,5 +1,3 @@
----
-
 elasticsearch7:
   image: docker.elastic.co/elasticsearch/elasticsearch:7.16.0
   environment:
--- a/toucan_connectors/awsathena/awsathena_connector.py
+++ b/toucan_connectors/awsathena/awsathena_connector.py
@@ -60,13 +60,19 @@ class AwsathenaConnector(ToucanConnector):
     region_name: str = Field(..., description='Your AWS region name')
 
     def get_session(self) -> boto3.Session:
-        return boto3.Session(
-            aws_access_key_id=self.aws_access_key_id,
-            # This is required because this gets appended by boto3
-            # internally, and a SecretStr can't be appended to an str
-            aws_secret_access_key=self.aws_secret_access_key.get_secret_value(),
-            region_name=self.region_name,
-        )
+        try:
+            aws_secret_access_key = self.aws_secret_access_key.get_secret_value()
+        except AttributeError as excp:
+            raise AttributeError(f'Error: failed to get your aws_secret_access_key, {excp}')
+        else:
+            boto_session = boto3.Session(
+                aws_access_key_id=self.aws_access_key_id,
+                # This is required because this gets appended by boto3
+                # internally, and a SecretStr can't be appended to an str
+                aws_secret_access_key=aws_secret_access_key,
+                region_name=self.region_name,
+            )
+            return boto_session
 
     @staticmethod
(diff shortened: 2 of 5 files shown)


tests/docker-compose.yml
```diff
@@ -1,3 +1,5 @@
+---
+
```
lukapeschke, line 2: why is this needed ?

toucan_connectors/mongo/mongo_connector.py
```diff
@@ -162,12 +162,19 @@ def get_status(self) -> ConnectorStatus:
...
+                client.server_info()
+            except pymongo.errors.ServerSelectionTimeoutError as e:
+                return ConnectorStatus(
+                    status=False, details=self._get_details(2, False), error=str(e)
+                )
+            except pymongo.errors.OperationFailure as e:
+                return ConnectorStatus(
+                    status=False, details=self._get_details(3, False), error=str(e)
+                )
+        except pymongo.errors.ConfigurationError as e:
```
lukapeschke, line 177: could you please unnest the `try` blocks ?

toucan_connectors/awsathena/awsathena_connector.py
```diff
@@ -60,13 +60,19 @@ class AwsathenaConnector(ToucanConnector):
...
+            raise AttributeError(f'Error: failed to get your aws_secret_access_key, {excp}')
+        else:
+            boto_session = boto3.Session(
+                aws_access_key_id=self.aws_access_key_id,
+                # This is required because this gets appended by boto3
+                # internally, and a SecretStr can't be appended to an str
+                aws_secret_access_key=aws_secret_access_key,
+                region_name=self.region_name,
+            )
+            return boto_session
```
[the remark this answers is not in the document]
Sanix-Darker: This is the error raised : 
```
AttributeError: 'NoneType' object has no attribute 'get_secret_value'
```
IMO, the message i added seems more explicit

tests/docker-compose.yml
Sanix-Darker: a result of my yarn linter 😅 !

lukapeschke: Hmmm that is only needed if you want to describe several YAML objects in the same file, such as for a kubernetes manifest... It's not needed here, so I'd keep the file as is

toucan_connectors/awsathena/awsathena_connector.py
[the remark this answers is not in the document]
lukapeschke: I agree that the message is better, but I'm not sure it should be possible to be in a state where we do not have a secret key. Do you know why it is the only optional field ?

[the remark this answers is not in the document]
Sanix-Darker: > Do you know why it is the only optional field ?

Nope 🤔 , i removed the "optional" !

toucan_connectors/mongo/mongo_connector.py
Sanix-Darker: done here : https://github.com/ToucanToco/toucan-connectors/pull/634/commits/c78eac60ab45a0e5ff6db630d67ba233c74491ed

tests/docker-compose.yml
Sanix-Darker: Thanks, did knew it 👍🏾

toucan_connectors/awsathena/awsathena_connector.py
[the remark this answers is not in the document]
lukapeschke: Since it has been removed, the `try/except` block is not needed anymore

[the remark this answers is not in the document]
Sanix-Darker: [`4af1caf` (#634)](https://github.com/ToucanToco/toucan-connectors/pull/634/commits/4af1caf91c6968e20bbe28be97e149e7b4eacd68)

PR #639, opened 2022-07-27
fix: Ensure STL dates are returned with the right dtype TCTC-3424

lukapeschke:
BigQuery and Athena translators store objects in the SQL DATE type as datetime.date objects.
That is correct, but pandas does not consider datetime.date objects as datetimes. As a result, when
serializing and deserializing a dataframe containing datetime.date objects, these get dumped/loaded
as strings.

This change ensures the right type is used.

TCTC-3424

## Checklist

* [x ] Unit tests for the changes exist
* [ x] Tests pass on CI and coverage remains at 100%
* [ x] Documentation reflects the changes where applicable

### DIFF
--- a/tests/utils/test_datetime.py
+++ b/tests/utils/test_datetime.py
@@ -1,0 +1,66 @@
+from datetime import date, datetime
+
+import pandas as pd
+import pytest
+from numpy import dtype
+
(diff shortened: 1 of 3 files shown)


raphaelvignes (approved):
Propre, thx!

PR #642, opened 2022-07-27
fix(mysql,athena) hidden table attr

raphaelvignes:
Old Athena & Mysql datasources config might have a Table attribute. 
We add them back but hidden from the UI to avoid breaking old configs.


lukapeschke (approved):
sweet

PR #645, opened 2022-08-03
fix(mongo-connector): search [TCTC-3545]

ninofiliu:
https://toucantoco.atlassian.net/browse/TCTC-3545

Searching for

```yml
or:
- foo: Foo
  bar: Bar
- baz: Baz
  qux: Qux
```

resulted in (pseudo mongo)

```yml
$or:
- foo: matches Foo
- bar: matches Bar
- baz: matches Baz
- qux: matches Qux
```

but it should result in (pseudo mongo)

```yml
$or:
- $and:
  - foo: matches Foo
  - bar: matches Bar
- $and:
  - baz: matches Baz
  - qux: matches Qux
```

It's more reasonable when reading the bug ticket: basically if I want to filter rows where name is John or Johnny and age is 20 I don't wanna be showned a John that is not 20 years old

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -10,7 +10,7 @@ multi_line_output = 3
 
 [tool.poetry]
 name = "toucan-connectors"
-version = "3.17.3"
+version = "3.17.4"
 description = "Toucan Toco Connectors"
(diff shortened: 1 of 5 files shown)


Albagnach (approved):
C'est ok pour moi.

PR #653, opened 2022-08-25
fix(mysql): Do not specify the database for schema-related functions TCTC-3768

lukapeschke:
## Change Summary

Both queries (SHOW DATABASES and the more complex one in build_database_model_extraction_query),
do not require a database to function.

Currently, the database was forced to "mysql", which caused the connection to fail in case the database
did not exist or if the SQL user used by the connector did not have sufficient permissions.

## Checklist

### DIFF
--- a/toucan_connectors/mysql/mysql_connector.py
+++ b/toucan_connectors/mysql/mysql_connector.py
@@ -99,9 +99,7 @@ class MySQLConnector(ToucanConnector, DiscoverableConnector):
         keep_untouched = (cached_property_with_ttl,)
 
     def _list_db_names(self) -> list[str]:
-        connection = pymysql.connect(
-            **self.get_connection_params(cursorclass=None, database='mysql')
-        )
+        connection = pymysql.connect(**self.get_connection_params(cursorclass=None, database=None))
         # Always add the suggestions for the available databases
         with connection.cursor() as cursor:
             cursor.execute('SHOW DATABASES;')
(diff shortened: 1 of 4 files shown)


toucan_connectors/mysql/mysql_connector.py
```diff
@@ -99,9 +99,7 @@ class Config:
         keep_untouched = (cached_property_with_ttl,)
 
     def _list_db_names(self) -> list[str]:
-        connection = pymysql.connect(
-            **self.get_connection_params(cursorclass=None, database='mysql')
-        )
+        connection = pymysql.connect(**self.get_connection_params(cursorclass=None, database=None))
```
Sanix-Darker, line 102: It could be a good idea to do the same thing as it's done on Postgres with a default database parameter, what do you think? 🤔 

https://github.com/ToucanToco/toucan-connectors/blob/master/toucan_connectors/postgres/postgresql_connector.py#L227

lukapeschke: Maybe in another PR, but we'd have to dig first, as Postgres and MySQL have different behaviours regarding metadata. Plus I don't want to introduce a side effect on a connector in a fix targetting MySQL

Sanix-Darker: Good point ! 👍🏽

Sanix-Darker (approved):

PR #658, opened 2022-08-30
deps: bumping some libs

Sanix-Darker:
## Change Summary

Bumping some libs :

- Explicit bumps :
  • Updating requests (2.27.1 -> 2.28.0)
  • Updating black (22.3.0 -> 22.6.0)
  • Updating pytest-mock (3.7.0 -> 3.8.2)
  • Updating pytest-asyncio (0.18.3 -> 0.19.0)
  • Updating werkzeug (2.1.2 -> 2.2.1)

- Implicit bumps (transient) :
  • Updating urllib3 (1.26.9 -> 1.26.11)
  • Updating attrs (21.4.0 -> 22.1.0)
  • Updating cffi (1.15.0 -> 1.15.1)
  • Updating numpy (1.23.0 -> 1.23.1)
  • Updating pandas (1.4.2 -> 1.4.3)
  • Updating coverage (6.4.1 -> 6.4.2)
  • Updating typing-extensions (4.2.0 -> 4.3.0)

## Checklist


lukapeschke (approved):
LGTM, although I believe we should definitely drop our dependency on Werkzeug, as we only use  it to throw a `BadRequest` exception in the databricks connector:

1. That's not the right exception type: The connector should not throw anything HTTP-related
2. It's a leaky abstraction: It was probably done like this because we wanted to have our API return a 400 error in case we can't start a cluster, but that's most probably not what other library users might want
3. It's a useless dependency **and** having such recent versions might conflict with other user's depencies (Id they're still on flask<2 for example)

PR #659, opened 2022-09-01
deps: remove werkzeug (BadRequest) dependency [TCTC-3805]

Sanix-Darker:
## Change Summary

- deps: bumping some libs
   - Implicit bumps:
     • Updating pandas (1.4.3 -> 1.4.4)
     • Updating pathspec (0.9.0 -> 0.10.0)

   - Explicit bumps:
     • Updating black (22.6.0 -> 22.8.0)
     • Updating pydantic (1.9.1 -> 1.10.1)

- feat: created a custom databricks exception + update tests
- deps: removed werkzeug from librairies

## Checklist


pyproject.toml
```diff
@@ -24,7 +24,7 @@ aiohttp = "^3.8.1"
 cached-property = "^1.5.2"
 Jinja2 = "^3.0.3"
 jq = "^1.2.2"
-pydantic = "^1.9.1"
+pydantic = "^1.10.1"
```
lukapeschke, line 27: I don't think pydantic should be bumped here if we don't need the update in the lib. Constraints should be as loose as possible for libs, else it can be annoying for the users

toucan_connectors/databricks/databricks_connector.py
```diff
@@ -6,14 +6,17 @@
...
 from requests.auth import HTTPBasicAuth
-from werkzeug.exceptions import BadRequest
 
 from toucan_connectors.common import ConnectorStatus, pandas_read_sql
 from toucan_connectors.toucan_connector import ToucanConnector, ToucanDataSource
 
 logger = logging.getLogger(__name__)
 
 
+class DatabricksClusterException(Exception):
```
lukapeschke, line 16: Could you please create a base Exception class for all errors than can occur on cluster start ? Otherwise we'll have to handle every single one of them explicitly

lukapeschke (changes requested):

pyproject.toml
Sanix-Darker: https://github.com/ToucanToco/toucan-connectors/pull/659/commits/c77f10c8b8686c5eb50205c4b3636ed9ccf97846

toucan_connectors/databricks/databricks_connector.py
Sanix-Darker: https://github.com/ToucanToco/toucan-connectors/pull/659/commits/06de879fc778a0de4365c1874f6b48bdaadabdc9

lukapeschke (approved):
Nice :100:

PR #662, opened 2022-09-04
feat: get target version per engine [TCTC-3498]

Sanix-Darker:
## Change Summary

We want to get the target source version for each engine, but on this Pull Request, we will be focusing on Mysql/PostgreSql and mongo.

- add an abstract method in a class called VersionableEngineConnector from toucan_connector
- set up target version handler for mysql + test
- set up target version handler for postgresql + test
- set up target version handler for mongo + test

## Checklist

### DIFF
--- a/tests/mongo/test_mongo.py
+++ b/tests/mongo/test_mongo.py
@@ -584,16 +584,18 @@ def test_status_unreachable(mongo_connector, mocker):
 
 
 def test_get_engine_version(mocker, mongo_connector):
+    # Should be a valide semver version converted to tuple
     mocker.patch('pymongo.MongoClient.server_info', return_value={'version': '3.4.5'})
-
     assert mongo_connector.get_engine_version() == (3, 4, 5)
 
+    # Should raise a MalformattedVersion error
     mocker.patch(
         'pymongo.MongoClient.server_info', return_value={'version': '--bad-version-format-'}
     )
     with pytest.raises(MalformattedVersion):
         assert mongo_connector.get_engine_version()
 
+    # Should raise an UnavailableVersion error
     mocker.patch('pymongo.MongoClient.server_info', return_value=None)
     with pytest.raises(UnavailableVersion):
         assert mongo_connector.get_engine_version()
--- a/tests/mongo/test_mongo.py
+++ b/tests/mongo/test_mongo.py
@@ -18,6 +18,7 @@ from toucan_connectors.mongo.mongo_connector import (
     _format_explain_result,
     normalize_query,
 )
+from toucan_connectors.toucan_connector import MalformattedVersion, UnavailableVersion
 
 
 @pytest.fixture(scope='module')
@@ -580,6 +581,22 @@ def test_status_unreachable(mongo_connector, mocker):
         ],
         error='qwe',
     )
+
+
+def test_get_engine_version(mocker, mongo_connector):
+    mocker.patch('pymongo.MongoClient.server_info', return_value={'version': '3.4.5'})
+
+    assert mongo_connector.get_engine_version() == (3, 4, 5)
+
+    mocker.patch(
+        'pymongo.MongoClient.server_info', return_value={'version': '--bad-version-format-'}
+    )
+    with pytest.raises(MalformattedVersion):
+        assert mongo_connector.get_engine_version()
+
+    mocker.patch('pymongo.MongoClient.server_info', return_value=None)
+    with pytest.raises(UnavailableVersion):
+        assert mongo_connector.get_engine_version()
 
 
 def test_status_bad_username(mongo_connector):
--- a/tests/mongo/test_mongo.py
+++ b/tests/mongo/test_mongo.py
@@ -18,7 +18,7 @@ from toucan_connectors.mongo.mongo_connector import (
     _format_explain_result,
     normalize_query,
 )
-from toucan_connectors.toucan_connector import MalformattedVersion, UnavailableVersion
+from toucan_connectors.toucan_connector import MalformedVersion, UnavailableVersion
 
 
 @pytest.fixture(scope='module')
@@ -588,11 +588,11 @@ def test_get_engine_version(mocker, mongo_connector):
     mocker.patch('pymongo.MongoClient.server_info', return_value={'version': '3.4.5'})
     assert mongo_connector.get_engine_version() == (3, 4, 5)
 
-    # Should raise a MalformattedVersion error
+    # Should raise a MalformedVersion error
     mocker.patch(
         'pymongo.MongoClient.server_info', return_value={'version': '--bad-version-format-'}
     )
-    with pytest.raises(MalformattedVersion):
+    with pytest.raises(MalformedVersion):
         assert mongo_connector.get_engine_version()
 
     # Should raise an UnavailableVersion error
--- a/toucan_connectors/mongo/mongo_connector.py
+++ b/toucan_connectors/mongo/mongo_connector.py
@@ -15,6 +15,8 @@ from toucan_connectors.toucan_connector import (
     DataStats,
     ToucanConnector,
     ToucanDataSource,
+    UnavailableVersion,
+    VersionableEngineConnector,
     decorate_func_with_retry,
     strlist_to_enum,
 )
@@ -89,7 +91,7 @@ class MongoDataSource(ToucanDataSource):
         return create_model('FormSchema', **constraints, __base__=cls).schema()
 
 
-class MongoConnector(ToucanConnector):
+class MongoConnector(ToucanConnector, VersionableEngineConnector):
     """Retrieve data from a [MongoDB](https://www.mongodb.com/) database."""
 
     data_source_model: MongoDataSource
@@ -338,6 +340,14 @@ class MongoConnector(ToucanConnector):
         data_source_rendered.query = normalize_query(data_source.query, data_source.parameters)
         del data_source_rendered.parameters
         return data_source_rendered.dict()
+
+    def get_engine_version(self) -> tuple:
+        client = pymongo.MongoClient(**self._get_mongo_client_kwargs())
+        try:
+            version = client.server_info()['version']
+            return super()._format_version(version)
+        except (TypeError, KeyError) as exc:
+            raise UnavailableVersion from exc
 
 
 def _format_explain_result(explain_result):
--- a/toucan_connectors/mysql/mysql_connector.py
+++ b/toucan_connectors/mysql/mysql_connector.py
@@ -347,5 +347,6 @@ class MySQLConnector(ToucanConnector, DiscoverableConnector, VersionableEngineCo
             except (TypeError, KeyError) as exc:
                 raise UnavailableVersion from exc
 
+
 class InvalidQuery(Exception):
     """raised when a query is invalid"""
--- a/toucan_connectors/mysql/mysql_connector.py
+++ b/toucan_connectors/mysql/mysql_connector.py
@@ -16,6 +16,8 @@ from toucan_connectors.toucan_connector import (
     TableInfo,
     ToucanConnector,
     ToucanDataSource,
+    UnavailableVersion,
+    VersionableEngineConnector,
     strlist_to_enum,
 )
 from toucan_connectors.utils.pem import InvalidPEMFormat, sanitize_spaces_pem
@@ -91,7 +93,7 @@ class SSLMode(str, Enum):
     VERIFY_CA = 'VERIFY_CA'
 
 
-class MySQLConnector(ToucanConnector, DiscoverableConnector):
+class MySQLConnector(ToucanConnector, DiscoverableConnector, VersionableEngineConnector):
     """
     Import data from MySQL database.
     """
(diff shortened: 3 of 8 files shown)


toucan_connectors/toucan_connector.py
```diff
@@ -472,6 +472,10 @@ def get_identifier(self):
     def describe(self, data_source: ToucanDataSource):
         """ """
 
+    def get_target_version(self) -> str | None:
+        """Tries to get the version of the target SGBD/source/service"""
+        return None
```
lukapeschke, line 477: I'd rather have a mixin for connectors that support retrieving the version of their target, such as `DiscoverableConnector`. That way, the signature could be changed to `Self -> str`, and we would not have to call the method to know if a connector actually supports it

toucan_connectors/mysql/mysql_connector.py
```diff
@@ -255,6 +255,19 @@ def _retrieve_data(self, datasource):
...
+        """
+        We try to get the MySQL version by running a query with our connection
+        """
+        connection = pymysql.connect(**self.get_connection_params())
+
+        with connection.cursor() as cursor:
+            cursor.execute('SELECT VERSION()')
+            version = cursor.fetchone()
+
+            # {'VERSION()': '8.0.30'}
```
lukapeschke, line 268:
```suggestion
```

tests/mongo/test_mongo.py
```diff
@@ -582,6 +583,11 @@ def test_status_unreachable(mongo_connector, mocker):
     )
 
 
+def test_get_version(mongo_connector):
+    # "in" and not "==", because sometimes, we can have "8.0.0.-blueeyes" as version
```
lukapeschke, line 587: I think the version should be cleaned by the connectors, as we will want strict comparisons later on. A semver regex will probably do the job for most connectors

tests/docker-compose.yml
```diff
@@ -17,7 +17,7 @@ elasticsearch8:
     - 9200:9200
 
 mongo:
-  image: mongo
+  image: mongo:6.0.1
```
lukapeschke, line 20: I'd rather have the DB clients mocked than pin minor versions here

tests/conftest.py
```diff
@@ -12,6 +12,12 @@
 
 from toucan_connectors.oauth2_connector.oauth2connector import SecretsKeeper
 
+ENGINE_VERSIONS = {
```
lukapeschke, line 15: should be a fixture, you're not supposed to import from conftest

lukapeschke (commented):
Thanks for tackling this! However, I think it is a bit early: We need to specifically define our need, and what WB expects (what format, maybe an integer tuple for easier comparisons) as an input format before starting the actual implementation here

toucan_connectors/toucan_connector.py
Sanix-Darker: Am not seeing how this can be done without calling a method 🤔, do you have an example please ?
Am okay with a class that need to be created and used just like DiscoverableConnector and i did that here : [`d0a9f06` (#662)](https://github.com/ToucanToco/toucan-connectors/pull/662/commits/d0a9f06262dc83541b99c93662e1f4bc7104faaf)

toucan_connectors/mysql/mysql_connector.py
Sanix-Darker: 👍🏽

toucan_connectors/mongo/mongo_connector.py
```diff
@@ -339,6 +341,14 @@ def _get_unique_datasource_identifier(self, data_source: MongoDataSource) -> dic
         del data_source_rendered.parameters
         return data_source_rendered.dict()
 
+    def get_engine_version(self) -> tuple:
+        client = pymongo.MongoClient(**self._get_mongo_client_kwargs())
+        try:
+            version = client.server_info()['version']
+            return super()._format_version(version)
```
Sanix-Darker, line 348:
```suggestion
        version = client.server_info()
        try:
            return super()._format_version(version['version'])
```

toucan_connectors/toucan_connector.py
```diff
@@ -548,7 +548,7 @@ def _validate(self, engine_version: str | float | None) -> bool:
         A small validation function for incoming version format
         """
         if engine_version is None:
-            raise UnavailableVersion('This engine has an unavailable version')
+            raise UnavailableVersion  # pragma: no cover
```
Sanix-Darker, line 551: In case of using `VersionableEngineConnector._validate()` seperately !

toucan_connectors/toucan_connector.py
```diff
@@ -504,3 +505,68 @@ def format_db_model(
...
+    """raise when the given version of the engine is not well formated"""
+
+
+class UnavailableVersion(Exception):
+    """raise when the version of the engine is not available"""
+
+
+class VersionableEngineConnector(ABC):
+    """
+    This class is responsable for formating and rendering the version engine
```
fspot, line 520:
```suggestion
    This class is responsible for formating and rendering the version engine
```

toucan_connectors/toucan_connector.py
lukapeschke: `isinstance(conn, VersionableEngineConnector)` rather than `if (version := conn.get_engine_version()) is not None`

toucan_connectors/toucan_connector.py
```diff
@@ -504,3 +505,68 @@ def format_db_model(
             .reset_index()
             .to_dict('records')
         )
+
+
+class MalformattedVersion(Exception):
```
lukapeschke, line 510:
```suggestion
class MalformedVersion(Exception):
```

toucan_connectors/toucan_connector.py
```diff
@@ -504,3 +505,68 @@ def format_db_model(
...
+    """raise when the given version of the engine is not well formated"""
```
lukapeschke, line 511:
```suggestion
    """raised when the given version of the engine is not well formated"""
```

toucan_connectors/toucan_connector.py
```diff
@@ -504,3 +505,68 @@ def format_db_model(
...
+
+
+class UnavailableVersion(Exception):
+    """raise when the version of the engine is not available"""
```
lukapeschke, line 515:
```suggestion
    """raised when the version of the engine is not available"""
```

toucan_connectors/toucan_connector.py
lukapeschke: It's a private method, so this is not supposed to happen

toucan_connectors/toucan_connector.py
```diff
@@ -504,3 +505,68 @@ def format_db_model(
...
+
+
+class VersionableEngineConnector(ABC):
+    """
+    This class is responsable for formating and rendering the version engine
+    from which the connector is fetching data from !
+
+    We suppose all versions should fit the semver format and for that, we
```
lukapeschke, line 523: During the meeting, we said that we do **not** suppose that all versions are semver. Matching logic looks good though :slightly_smiling_face:

toucan_connectors/toucan_connector.py
```diff
@@ -504,3 +505,68 @@ def format_db_model(
...
+    2- Minor (optional - requires major to match)
+    3- Patch (optional - requires minor to match)
+    4- Meta/pre-release (optional - requires minor to match)
+    You can add the ^ prefix and $suffix to do match enforcement.
+    """
+
+    # The output of these rules :
+    semver_regex = (
+        r'(0|(?:[1-9]\d*))(?:\.(0|(?:[1-9]\d*))(?:\.(0|(?:[1-9]\d*)))?(?:\-([\w][\w\.\-_]*))?)?'
+    )
```
lukapeschke, line 544:
```suggestion
    semver_regex = re.compile(
        r'(0|(?:[1-9]\d*))(?:\.(0|(?:[1-9]\d*))(?:\.(0|(?:[1-9]\d*)))?(?:\-([\w][\w\.\-_]*))?)?'
    )
```

toucan_connectors/toucan_connector.py
```diff
@@ -504,3 +505,68 @@ def format_db_model(
...
+
+    def _validate(self, engine_version: str | float | None) -> bool:
+        """
+        A small validation function for incoming version format
+        """
+        if engine_version is None:
+            raise UnavailableVersion  # pragma: no cover
+
+        return re.match(self.semver_regex, str(engine_version)) is not None
```
lukapeschke, line 553:
```suggestion
        return self.semver_regex.match(str(engine_version)) is not None
```

toucan_connectors/toucan_connector.py
```diff
@@ -504,3 +505,68 @@ def format_db_model(
...
+        """
+        We get the input version from the engine and convert it in tuple
+        Either :
+            - 1.45 -> (1, 45)
+            - "0.9.3" -> (0, 9, 3)
+            - "3.4.57 Debian+only" -> (3, 4, 57)
+        """
+        if self._validate(input_version):
+            return tuple(
+                [int(x) for x in re.findall(self.semver_regex, str(input_version))[0] if len(x)]
```
lukapeschke, line 569:
```suggestion
                int(x) for x in self.semver_regex.findall(str(input_version))[0] if len(x)
```

fspot (approved):
lgtm :1st_place_medal:

toucan_connectors/toucan_connector.py
```diff
@@ -504,3 +505,68 @@ def format_db_model(
...
+
+    def _format_version(self, input_version: str | float | None = None) -> tuple:
+        """
+        We get the input version from the engine and convert it in tuple
+        Either :
+            - 1.45 -> (1, 45)
+            - "0.9.3" -> (0, 9, 3)
+            - "3.4.57 Debian+only" -> (3, 4, 57)
+        """
+        if self._validate(input_version):
```
lukapeschke, line 567: We're matching the same thing twice here... Maybe `_validate` could return `re.Match | None` ?

toucan_connectors/postgres/postgresql_connector.py
```diff
@@ -238,3 +240,17 @@ def _list_tables_info(self, database_name: str = None) -> List[tuple]:
...
             return cursor.fetchall()
+
+    def get_engine_version(self) -> tuple:
+        """
+        We try to get the PostgreSQL version by running a query with our connection
+        """
+        connection = pgsql.connect(**self.get_connection_params(database=self.default_database))
+
+        with connection.cursor() as cursor:
+            cursor.execute("select current_setting('server_version');")
```
lukapeschke, line 251:
```suggestion
            cursor.execute("SELECT CURRENT_SETTING('server_version');")
```

toucan_connectors/toucan_connector.py
```diff
@@ -504,3 +505,68 @@ def format_db_model(
...
+            return tuple(
+                [int(x) for x in re.findall(self.semver_regex, str(input_version))[0] if len(x)]
+            )
+
+        raise MalformattedVersion(f'"{input_version}" is not a valid semver version')
```
lukapeschke, line 572:
```suggestion
        raise MalformattedVersion(f'"{input_version}" is not a valid understood')
```

We want to support more than semver

toucan_connectors/toucan_connector.py
```diff
@@ -504,3 +505,68 @@ def format_db_model(
...
+        if engine_version is None:
+            raise UnavailableVersion  # pragma: no cover
+
+        return re.match(self.semver_regex, str(engine_version)) is not None
+
+    @abstractmethod
+    def get_engine_version(self) -> tuple:
+        """This method will return the version of the target engine"""
+
+    def _format_version(self, input_version: str | float | None = None) -> tuple:
```
lukapeschke, line 559: Could you please add a unit test with a lot of different inputs for this ? Like `1`, `'1'`, `1.2`, `'1.2'`, `'14.3 (Debian 14.3-1.pgdg110+1)'` etc ?

lukapeschke (changes requested):
Thanks for the update! New implem looks good :medal_sports: A few inline comments

toucan_connectors/toucan_connector.py
Sanix-Darker: `is not a valid understood` ? what about `is not a valid version` ? 🤔

lukapeschke: sorry, I mean `is not understood`. `Is not a valid version` is not quite true, as the version is valid from the PoV of the engine, it's us who can't interpret it properly

toucan_connectors/toucan_connector.py
Sanix-Darker: [`62b6294` (#662)](https://github.com/ToucanToco/toucan-connectors/pull/662/commits/62b629407cf559a8e539308321e3d7a513f68afb)

toucan_connectors/toucan_connector.py
Sanix-Darker: [`5011580` (#662)](https://github.com/ToucanToco/toucan-connectors/pull/662/commits/50115806b5659d1d76dfe2349071565062fff3da)

lukapeschke (approved):

PR #669, opened 2022-09-13
fix: bigQuery variables interpolation [TCTC-3642]

Sanix-Darker:
## Change Summary

It seems like, when a give variable has a space between '@' and its keys
name '__front_var_0__', it's impossible for bigquery to set well that
variable, since it's based on `str.find` (on _prepare_parameters function) as for an example :

```python
>>> query = 'SELECT @ __front_var_0__';
>>> query.find('@' + '__front_var_0');
-1
>>> query.find('@ ' + '__front_var_0__');
12
```
Screenshot from 2022-09-13 08-58-36

## Checklist


toucan_connectors/google_big_query/google_big_query_connector.py
```diff
@@ -169,7 +169,9 @@ def _prepare_parameters(query: str, parameters: Optional[Dict]) -> List:
     @staticmethod
     def _prepare_query(query: str) -> str:
         """replace ToucanToco variable definition by Google Big Query variable definition"""
-        new_query = query.replace('{{', '@').replace('}}', '')
+        new_query = (
+            query.replace('{{ ', '{{').replace('{{', '@').replace(' }}', '}}').replace('}}', '')
+        )
```
jGundermann, line 174: There is a helper function for the interpolation named : sanitize_query.
You can have a example here 
https://github.com/ToucanToco/toucan-connectors/blob/master/toucan_connectors/awsathena/awsathena_connector.py#L58

jGundermann (changes requested):

jGundermann (approved):
LGTM

Sanix-Darker:
So this should be renamed as "feat: bigQuery, add no injection SQL to interpolated variables"

Sanix-Darker:
Confirmed that it's now working, so, i can merge it !
Thanks for the review @jGundermann

PR #693, opened 2022-09-27
fix(ci): Pin python version to 3.10.6

lukapeschke:


lukapeschke:
Oops, there's no mypy here :facepalm:

PR #701, opened 2022-09-28
fix(utils): Pass utc=True in sanitize_df_dates util

lukapeschke:
pd.to_datetime() creates np.datetime64 objects, which are naive. Thus, in order to be converted to UTC, we must pass utc=True.

relates to TCTC-4151

## Checklist


fspot (approved):
:v:

PR #706, opened 2022-10-03
fix: sql injection posgresql [TCTC-3809]

Sanix-Darker:
## WHAT
- feat: added sanitizer on postgresql query to prevent sql injection
- fix: small typo on athena variable form


Sanix-Darker:
😵‍💫 oups HS !

PR #707, opened 2022-10-04
fix: dict as a sequence when we pass % instead of %% [TCTC-3809]

Sanix-Darker:
## Change Summary

- Well formating of sql-query to prevent a dict is not a sequence when passing % instead of %%


tests/mysql/test_mysql.py
```diff
@@ -200,6 +200,22 @@ def test_get_df(mocker: MockerFixture):
     mysql_connector.get_df(data_source)
     reasq.assert_called_once_with('select * from Country', con=snock(), params={})
 
+    # With query havinf %
```
Sanix-Darker, line 203:
```suggestion
    # With query having %
```

lukapeschke (approved):

PR #715, opened 2022-10-07
fix/alphanumeric percentage only

Sanix-Darker:
## Change Summary

We should replace % to %% only for alphanumeric characters to prevent breaking variables

- feat: we should handle only alpha numerics elements
- test: added a test accordingly

### DIFF
--- a/toucan_connectors/common.py
+++ b/toucan_connectors/common.py
@@ -404,7 +404,7 @@ def pandas_read_sql(
         # FIXME: We should use here the sqlalchemy.text module to
         # escape characters like % but as a quick fix, we left replace
         query = query.replace('%%', '%')
-        query = re.sub(r'%[A-Za-z0-9]+%', r'%\g<0>%', query)
+        query = re.sub(r'%[^%()]+%', r'%\g<0>%', query)
         df = pd.read_sql(query, con=con, params=params, **kwargs)
     except pd.io.sql.DatabaseError:
         if is_interpolating_table_name(query):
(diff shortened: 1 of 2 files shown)


toucan_connectors/common.py
```diff
@@ -404,7 +404,7 @@ def pandas_read_sql(
         # FIXME: We should use here the sqlalchemy.text module to
         # escape characters like % but as a quick fix, we left replace
         query = query.replace('%%', '%')
-        query = re.sub(r'%[^%]+%', r'%\g<0>%', query)
+        query = re.sub(r'%[A-Za-z0-9]+%', r'%\g<0>%', query)
```
fspot, line 407: this is too specific (we cannot do `LIKE '%foo bar%'` for example)
we should just exclude patterns like `%(xxx)s` IMO

Sanix-Darker: true, it's better 🙏🏽

toucan_connectors/common.py
```diff
@@ -404,7 +404,7 @@ def pandas_read_sql(
         # FIXME: We should use here the sqlalchemy.text module to
         # escape characters like % but as a quick fix, we left replace
         query = query.replace('%%', '%')
-        query = re.sub(r'%[^%]+%', r'%\g<0>%', query)
+        query = re.sub(r'%[^%()]+%', r'%\g<0>%', query)
```
fspot, line 407:
```suggestion
        query = re.sub(r'%[^(]', r'%\g<0>', query)
```

fspot (approved):

PR #722, opened 2022-10-18
fix(common): Add support for __VOID__ to nosql_apply_parameters_to_query TCTC-4267

lukapeschke:


lukapeschke:
@Sanix-Darker FYI the last commit is because when building a VQB pipeline, we receive variables properly, but in case of a story not being edited, the front interpolates them! So we also need to remove any query value with a `__VOID__` (and fix this in toucan v3 :sweat: )

Sanix-Darker:
> @Sanix-Darker FYI the last commit is because when building a VQB pipeline, we receive variables properly, but in case of a story not being edited, the front interpolates them! So we also need to remove any query value with a __VOID__ (and fix this in toucan v3 sweat )

ouch...

toucan_connectors/mongo/mongo_connector.py
```diff
@@ -24,9 +24,37 @@
...
+
+def _is_match_statement(d: Any) -> bool:
+    return isinstance(d, dict) and list(d.keys()) == ['$match']
+
+
+def _remove_empty_matches_from_query(query: dict | list[dict]) -> Any:
+    """Removes empty $match operations"""
+    if isinstance(query, list):
+        return [q for q in query if not (_is_match_statement(q) and _is_match_empty(q))]
+    return query
```
Sanix-Darker, line 52: :eyes-bleeding:

toucan_connectors/mongo/mongo_connector.py
```diff
@@ -24,9 +24,37 @@
...
+
+
 def normalize_query(query, parameters):
     query = nosql_apply_parameters_to_query(query, parameters)
 
+    # FIXME: This removes empty $match operations that could have been created while removing
+    # missing parameters from the query in nosql_apply_parameters to query. However, this way of
+    # handling __VOID__ filters is hacky and should be implemented earlier, when translating the VQB
+    # pipeline
+    query = _remove_empty_matches_from_query(query)
```
Sanix-Darker: +100000

Sanix-Darker (commented):
LGTM !
But it could be really good for an old school to have an overview on this for a large-scale impact !

cc @fspot @davinov ? 🙏🏽

fspot (approved):
:+1:

PR #740, opened 2022-10-28
fix(postgresql): Ensure we always use default_database if provided TCTC-4462

lukapeschke:
## Change Summary

Ensure we use the connector's default database rather than `'postgres'` when checking connection status

## Checklist


jGundermann (approved):
LGTM

PR #771, opened 2022-11-15
refactor(snowflake): Simplify the snowflake connector TCTC-4601

lukapeschke:
Work items:

* Stop using SnowflakeCommon
* Stop using ThreadPools
* Stop using connection pools
* Add mypy checks

TCTC-4601

### DIFF
--- a/tests/http_api/test_http_api.py
+++ b/tests/http_api/test_http_api.py
@@ -481,7 +481,7 @@ def test_get_cache_key(connector, auth, data_source):
     data_source.parameters = {'first_name': 'raphael'}
     key = connector.get_cache_key(data_source)
 
-    assert key == 'b35e0189-a527-3f9c-827d-b0c54e23b780'
+    assert key == 'fa95c942-9b94-3f07-9ed4-24c34abfbdae'
 
     data_source.headers = {'name': '{{ first_name }}'}  # change the templating style
     key2 = connector.get_cache_key(data_source)
--- a/tests/test_connector.py
+++ b/tests/test_connector.py
@@ -163,7 +163,7 @@ def test_get_cache_key():
     key = connector.get_cache_key(ds)
     # We should get a deterministic identifier:
     # /!\ the identifier will change if the model of the connector or the datasource changes
-    assert key == '3ffe037e-3559-3a17-aede-c4ab0669cc25'
+    assert key == '10ae8360-5e30-319e-8cb8-a8a817826d12'
 
     ds.query = 'wow'
     key2 = connector.get_cache_key(ds)
--- a/toucan_connectors/snowflake/snowflake_connector.py
+++ b/toucan_connectors/snowflake/snowflake_connector.py
@@ -334,6 +334,7 @@ class SnowflakeConnector(ToucanConnector[SnowflakeDataSource], DiscoverableConne
         warehouse: str | None = None,
         database: str | None = None,
         as_df: Literal[True] = ...,
+        snowflake_connection: SnowflakeConnection | None = None,
     ) -> pd.DataFrame:
         ...  # pragma: no cover
 
@@ -346,6 +347,7 @@ class SnowflakeConnector(ToucanConnector[SnowflakeDataSource], DiscoverableConne
         warehouse: str | None = None,
         database: str | None = None,
         as_df: Literal[False],
+        snowflake_connection: SnowflakeConnection | None = None,
     ) -> list[dict]:
         ...  # pragma: no cover
 
@@ -357,14 +359,20 @@ class SnowflakeConnector(ToucanConnector[SnowflakeDataSource], DiscoverableConne
         warehouse: str | None = None,
         database: str | None = None,
         as_df: bool = True,
+        snowflake_connection: SnowflakeConnection | None = None,
     ) -> pd.DataFrame | list[dict]:
-        with self._get_connection(database=database, warehouse=warehouse) as conn:
+        def _execute(conn: SnowflakeConnection) -> pd.DataFrame | list[dict]:
             curs = conn.cursor(SfDictCursor)
             query_result = curs.execute(query, parameters)
             assert query_result is not None
             # snowflake typing is incomplete for DictCursor
             results: list[dict] = query_result.fetchall()  # type:ignore[assignment]
             return pd.DataFrame(results) if as_df else results
+
+        if snowflake_connection is not None:
+            return _execute(snowflake_connection)
+        with self._get_connection(database=database, warehouse=warehouse) as conn:
+            return _execute(conn)
 
     def _describe_query(self, query: str) -> dict[str, str]:
         with self._get_connection() as conn:
@@ -398,7 +406,12 @@ class SnowflakeConnector(ToucanConnector[SnowflakeDataSource], DiscoverableConne
         prepared_query, prepared_params = SqlQueryHelper.prepare_limit_query(
             data_source.query, data_source.parameters, offset=offset, limit=limit
         )
-        return self._execute_query(prepared_query, prepared_params)
+        return self._execute_query(
+            prepared_query,
+            prepared_params,
+            database=data_source.database,
+            warehouse=data_source.warehouse,
+        )
 
     def _retrieve_data(self, data_source: SnowflakeDataSource) -> pd.DataFrame:
         return self._fetch_data(data_source)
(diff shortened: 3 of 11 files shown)


toucan_connectors/snowflake/snowflake_connector.py
```diff
@@ -282,174 +306,134 @@ def _refresh_oauth_token(self):
...
+
+    def _describe_query(self, query: str) -> dict[str, str]:
+        with self._get_connection() as conn:
+            curs = conn.cursor(SfDictCursor)
+            return {r.name: _TYPE_CODE_MAPPING[r.type_code] for r in curs.describe(query)}
+
+    def _get_warehouses(self, warehouse_name: str | None = None) -> list[str]:
+        query = 'SHOW WAREHOUSES'
+        if warehouse_name:
+            query += f"LIKE '{warehouse_name}'"
```
fspot, line 377: there's a space missing, this will give `WAREHOUSESLIKE`

toucan_connectors/snowflake/snowflake_connector.py
```diff
@@ -282,174 +306,134 @@ def _refresh_oauth_token(self):
...
+
+        result = self._execute_query(query, warehouse=warehouse_name)
+        return result['name'].to_list() if 'name' in result.columns else []
+
+    def _get_databases(self, database_name: str | None = None) -> list[str]:
+        query = 'SHOW DATABASES'
+        if database_name:
+            query += f"LIKE '{database_name}'"
```
fspot, line 385: same here

toucan_connectors/snowflake/snowflake_connector.py
```diff
@@ -282,174 +306,134 @@ def _refresh_oauth_token(self):
...
+        limit: int | None = None,
+        get_row_count: bool | None = False,
     ) -> DataSlice:
-        data_source = self._set_warehouse(data_source)
-        with self._get_connection(data_source.database, data_source.warehouse) as connection:
-            result = SnowflakeCommon().get_slice(
-                connection, data_source, offset=offset, limit=limit, get_row_count=get_row_count
-            )
-        return result
+        # FIXME: are permissions ignored for the snowflake connector ?
```
fspot, line 414: Apparently yes − but not with the new way of applying permissions (as FilterSteps) so maybe we can ignore it ? or copy the code from `toucan_connector.py` otherwise

toucan_connectors/snowflake/snowflake_connector.py
```diff
@@ -282,174 +306,134 @@ def _refresh_oauth_token(self):
...
+        df = self._fetch_data(data_source, offset=offset, limit=limit)
+        return DataSlice(df=df, stats=DataStats(total_returned_rows=len(df)))
```
fspot, line 416: à remplacer par pagination infos quand l'autre PR sera mergée

fspot (changes requested):
J'ai review un peu en diagonale mais gg ça m'a l'air top, mes yeux ont juste spotté 1 bug

toucan_connectors/snowflake/snowflake_connector.py
lukapeschke: Woops good catch! Anyway, I want to test this with a real snowflake deployement before merging, since we're accessing the `"name"` column of a DF, and Snowflake might return `NAME`.

toucan_connectors/snowflake/snowflake_connector.py
lukapeschke: That's true! I'd rather leave it as is, since they were also ignored before

toucan_connectors/toucan_connector.py
```diff
@@ -101,7 +104,8 @@ class Config:
         validate_assignment = True
 
     @classmethod
-    def get_form(cls, connector: 'ToucanConnector', current_config):
+    # def get_form(cls, connector: Type['ToucanConnector'], current_config):
```
Sanix-Darker, line 107: why maintaining this comment ?

toucan_connectors/toucan_connector.py
```diff
@@ -252,9 +256,15 @@ def get_connector_secrets_form(cls) -> Optional[ConnectorSecretsForm]:
     """
     if hasattr(cls, 'get_connector_secrets_form'):
         return getattr(cls, 'get_connector_secrets_form')()
+    return None
+
 
+_UI_HIDDEN: dict[str, Any] = {'ui.hidden': True}
```
Sanix-Darker, line 262: great to have this.... question, is it not better to have it as a global variable that all connectors should import when they have fields with 'ui.hidden' ?

toucan_connectors/toucan_connector.py
```diff
@@ -61,36 +60,40 @@ class DataSlice(NamedTuple):
...
     the value is either <default value> or a tuple ( <type>, <default value> )
     If the field is required, the <default value> has to be '...' (cf pydantic doc)
     By default, the field is considered required.
     """
-    return StrEnum(field, {v: v for v in strlist}), default_value
+    return StrEnum(field, {v: v for v in strlist}), default_value  # type: ignore[call-overload]
 
 
-class ToucanDataSource(BaseModel):
+# Binding C to ToucanConnector causes mypy to crash (last checked with mypy 0.991)
```
Sanix-Darker, line 84: Is there a referal url for this issue yet ? if yes, providing the link in this comment could be great IMO !

toucan_connectors/snowflake/snowflake_connector.py
```diff
@@ -152,44 +171,45 @@ def schema_extra(schema: Dict[str, Any], model: Type['SnowflakeConnector']) -> N
...
             self.sso_credentials_keeper
             and self.sso_credentials_keeper.client_secret.get_secret_value()
         )
 
     @staticmethod
-    def _get_status_details(index: int, status: Optional[bool]):
+    def _get_status_details(index: int, status: bool | None) -> list[tuple[str, bool | None]]:
         checks = ['Connection to Snowflake', 'Default warehouse exists']
-        ok_checks = [(check, True) for i, check in enumerate(checks) if i < index]
+        ok_checks: list[tuple[str, bool | None]] = [
```
Sanix-Darker, line 195:
```suggestion
        ok_checks: list[tuple[str, bool]] = [
```

For what  can see, it's always True, never None

toucan_connectors/snowflake/snowflake_connector.py
```diff
@@ -152,44 +171,45 @@ def schema_extra(schema: Dict[str, Any], model: Type['SnowflakeConnector']) -> N
...
+            (check, True) for i, check in enumerate(checks) if i < index
+        ]
         new_check = (checks[index], status)
-        not_validated_checks = [(check, None) for i, check in enumerate(checks) if i > index]
+        not_validated_checks: list[tuple[str, bool | None]] = [
```
Sanix-Darker, line 199:
```suggestion
        not_validated_checks: list[tuple[str, None]] = [
```

toucan_connectors/snowflake/snowflake_connector.py
```diff
@@ -152,44 +171,45 @@ def schema_extra(schema: Dict[str, Any], model: Type['SnowflakeConnector']) -> N
...
+        ]
         return ok_checks + [new_check] + not_validated_checks
 
     def get_status(self) -> ConnectorStatus:
         try:
             res = self._get_warehouses(self.default_warehouse)
             if len(res) == 0:
-                raise SnowflakeConnectorWarehouseDoesNotExists(
-                    f"The warehouse '{self.default_warehouse}' does not exist."
+                return ConnectorStatus(
```
Sanix-Darker, line 208: 👍🏽

tests/http_api/test_http_api.py
```diff
@@ -481,7 +481,7 @@ def test_get_cache_key(connector, auth, data_source):
     data_source.parameters = {'first_name': 'raphael'}
     key = connector.get_cache_key(data_source)
 
-    assert key == 'b35e0189-a527-3f9c-827d-b0c54e23b780'
+    assert key == 'fa95c942-9b94-3f07-9ed4-24c34abfbdae'
```
Sanix-Darker, line 484: IMO... this is a flaky test, we should probably change this test... that's not the first time i had to update this value... not a blocking point tho... WDYT ?

tests/test_connector.py
```diff
@@ -163,7 +163,7 @@ def test_get_cache_key():
     key = connector.get_cache_key(ds)
     # We should get a deterministic identifier:
     # /!\ the identifier will change if the model of the connector or the datasource changes
-    assert key == '3ffe037e-3559-3a17-aede-c4ab0669cc25'
+    assert key == '10ae8360-5e30-319e-8cb8-a8a817826d12'
```
Sanix-Darker, line 166: same here... ?

toucan_connectors/toucan_connector.py
lukapeschke: woops, I'll remove it

toucan_connectors/toucan_connector.py
lukapeschke: I dont think so, as mypy is only running on this file and on snowflake_connector. Maybe someday if the pattern is repeated in too many places

toucan_connectors/toucan_connector.py
lukapeschke: Nope, guess mutually-referencing GATs is an edge case for mypy :sweat_smile:

toucan_connectors/snowflake/snowflake_connector.py
lukapeschke: Yes, but we need a looser type here since were concatenating all checks lists in the end

toucan_connectors/snowflake/snowflake_connector.py
lukapeschke: Same here, we need a looser type

tests/http_api/test_http_api.py
lukapeschke: I think the test is valid, it's probably the `get_cache_key` method that should be changed to not depend on a datasource model

fspot (approved):

PR #800, opened 2022-12-02
fix: google-sheet dates load double to string [TCTC-4148]

Sanix-Darker:
## Change Summary

Updated the `dateTimeRenderOption` parameter from *SERIAL_NUMBER* to *FORMATTED_STRING* to load dates as simple string, see : https://developers.google.com/sheets/api/reference/rest/v4/DateTimeRenderOption
*EDIT1*: Added a new param to let the user trigger that option as default, see : https://github.com/ToucanToco/toucan-connectors/pull/800#issuecomment-1334930144 and https://github.com/ToucanToco/toucan-connectors/pull/800#issuecomment-1334971509

## DATASET
Screenshot from 2022-12-02 09-53-28

### BEFORE
Screenshot from 2022-12-02 01-21-56

### AFTER
Screenshot from 2022-12-02 01-18-40

PS: to be honest... am wondering if we really have to do this change since the date infering will be "broken"... seems to be a product question tho... ?

## TODO ?

If the fix seems to be okay, it should be cherry-pick on V3 too.

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### Changed
+
+- Feat: The connector `GoogleSheets` datasource now has an option called `Render DateTime`, to trigger date time rendering when reading a sheet dataset.
+
 ### [3.23.5] 2022-11-07
 
 ### Changed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 ### Changed
 
-- Feat: The connector `GoogleSheets` datasource now has an option called `Render DateTime`, to trigger date time rendering when reading a sheet dataset.
+- Feat: The connector `GoogleSheets` datasource now has an option called `Dates as Floats`, to see date time columns as strings or float when reading the sheet.
 
 ### [3.23.5] 2022-11-07
 
--- a/toucan_connectors/google_sheets/google_sheets_connector.py
+++ b/toucan_connectors/google_sheets/google_sheets_connector.py
@@ -30,8 +30,8 @@ class GoogleSheetsDataSource(ToucanDataSource):
     header_row: int = Field(
         0, title='Header row', description='Row of the header of the spreadsheet'
     )
-    render_date_time: bool = Field(
-        False, title='Render DateTime', description='Render DateTime columns from the sheet'
+    dates_as_float: bool = Field(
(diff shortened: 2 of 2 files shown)


lukapeschke:
What do you mean by "date inference will be broken" ?

Sanix-Darker:
> What do you mean by "date inference will be broken" ?

maybe the word "broken" is wrong here, what i meant is that the date interpretation will not be the same as before, as you can see on pictures...

lukapeschke:
Right. I'd be in favor of adding an option defaulting to the new format to the datasource + an entry to the changelog. After that, we're good to go :slightly_smiling_face:

Sanix-Darker:
DEMO with the new option :
Peek 2022-12-02 10-20

lukapeschke:
The behaviour is the opposite of what I'd expect: When "render date time" is ticked, you get a float. When it's unticked, you get a datetime. Maybe a dropdown with `Literal["SERIAL_NUMBER", "FORMATTED_STRING"]` and a link to the doc in the tooltip would be the simplest ? Otherwise you can leave it as is, but maybe call the checkbox "Dates as floats" (unticked by default) to be more explicit ?

Sanix-Darker:
> The behaviour is the opposite of what I'd expect: When "render date time" is ticked, you get a float

that's what i did 🤔 ... or am not following something well ?
- checked (ticked):  dates as float
- unchecked (unticked): dates as String (default behaviour)


>  Maybe a dropdown with Literal["SERIAL_NUMBER", "FORMATTED_STRING"]

For the dropdown idea... IMO it's just two options and am not sure the AB need to know what it means "SERIAL_NUMBER"... etc... ? idk... he just want to check/uncheck ?

lukapeschke:
> > The behaviour is the opposite of what I'd expect: When "render date time" is ticked, you get a float
> 
> that's what i did thinking ... or am not following something well ?

In the demo, when you click on "run preview" with "render datetime" ticked, you get a float in response. It matches the code: 
```python
# If "render_date_time" is ticket, you'll select "SERIAL_NUMBER"
'SERIAL_NUMBER' if data_source.render_date_time else 'FORMATTED_STRING'
```

> For the dropdown idea... IMO it's just two options and am not sure the AB need to know what it means "SERIAL_NUMBER"... etc... ? idk... he just want to check/uncheck ?

As I said, I'm fine with both forms, as long as the behaviour is explicit.

lukapeschke (approved):

PR #810, opened 2022-12-05
Revert "build(deps-dev): bump pytest-rerunfailures from 10.2 to 10.3"

lukapeschke:
Reverts ToucanToco/toucan-connectors#809


lukapeschke:
This is not a patch release, and we're on the LTS branch

Sanix-Darker (approved):
mhhhh alright, i know we should not bump on LTS rn, i appproved/merged because it was a 'testing bump' that doesn't affect anything on our prod code 🤔 + i saw some fixes...


>- Fix crash when pytest-xdist is installed but disabled. (Thanks to @mgorny <https://github.com/mgorny>_ for the PR.)
>- Fix crash when xfail(strict=True) mark is used with --rerun-only flag.

lukapeschke:
Yes but we don't depend on pytest-xdist. Apart from patch releases or releases containing security-critical patches (like `cryptography`) , nothing should be merged in this branch

PR #838, opened 2022-12-21
feat: created a peakina connector

Sanix-Darker:
## WHAT

We are working on moving all our datasource to queries atm !
To achieve this, we need to have a kind of 'peakina connector' from toucan-connectors

## HOW

- feat: added peakina as a dependency in toucan-connectors
- feat: added a peakina connector with as datasource, the given peakina.DataSource
- test: added some tests ?

## SCREENSHOTS
Screenshot from 2022-12-21 17-29-29


toucan_connectors/peakina/peakina_connector.py
```diff
@@ -0,0 +1,21 @@
...
+
+from toucan_connectors.toucan_connector import ToucanConnector
+
+
+class PeakinaDataSource(DataSource):
+    class Config:
+        extra = 'allow'
+
+    def __init__(self, **data: Any) -> None:
+        super().__init__(**{k: v for k, v in data.items()})
```
lukapeschke, line 14:
```suggestion
        super().__init__(**data)
```

pyproject.toml
```diff
@@ -64,6 +64,7 @@ zeep = {version = "^4.1.0", optional = true}
 snowflake-connector-python = {version = "^2.7.12", optional = true}
 pyarrow = {version = "<9", optional = true}
 toucan-client = {version = "^1.0.1", optional = true}
+peakina = {version = "^0.9.1", extras = ["all"]}
```
lukapeschke, line 67: I think that for consistency, this should also be an optional extra

lukapeschke (commented):
LGTM overall, a few inline comments. Also, could you please enable mypy on the new connector ?

pyproject.toml
Sanix-Darker: https://github.com/ToucanToco/toucan-connectors/pull/838/commits/3f306c36c0203867952200ef48ac0e1799d93765

pyproject.toml
```diff
@@ -171,6 +172,9 @@ all = [
     "zeep"
 ]
 
+[tool.poetry.group.extras.dependencies]
+peakina = "^0.9.1"
```
lukapeschke, line 176: this should be in the `[tool.poetry.dev-dependencies]` group to avoid mixing up two config formats. Also, it needs to be included in the `all` extra

lukapeschke (changes requested):
Please enable mypy checks for the new connector

pyproject.toml
Sanix-Darker: [`377de19` (#838)](https://github.com/ToucanToco/toucan-connectors/pull/838/commits/377de191e4a897997bd3babd18a43340e6a742e0)

lukapeschke (approved):

PR #894, opened 2023-01-23
fix(filters): handle __VOID__ in $and match conditions

aklom:
## Change Summary
When there is a void match of a column inside an "and" condition, it is not filtered from the query, resulting to empty results of data. 

## Related issue number
x

## Checklist

### DIFF
--- a/toucan_connectors/mongo/mongo_connector.py
+++ b/toucan_connectors/mongo/mongo_connector.py
@@ -34,12 +34,14 @@ def _is_empty_match_column(elem: Any):
 
     return False
 
+
 def _is_match_empty(match_: dict) -> bool:
     return all(_is_empty_match_column(v) for v in match_.values())
 
 
 def _is_match_statement(d: Any) -> bool:
     return isinstance(d, dict) and list(d.keys()) == ['$match']
+
 
 def _sanitize_match(query: dict) -> dict:
     if _is_match_empty(query):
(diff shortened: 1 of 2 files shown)


toucan_connectors/mongo/mongo_connector.py
```diff
@@ -24,23 +24,35 @@
...
-        # empty
-        if elem == {}:
-            return True
-        if isinstance(elem, dict):
-            return _is_match_empty(elem)
-        return False
+def _is_empty_match_column(elem: Any):
+    # Not matching None here, because we don't want to consider {'$match': {'col': None}} to be
+    # empty
+    if elem == {}:
```
Sanix-Darker, line 30:
```suggestion
    if not bool(elem):
```

```bash
In [1]: not bool({})
Out[1]: True

In [2]: not bool([])
Out[2]: True
```
WDYT ?
```

Comment (aklom) [reply toucan_connectors/mongo/mongo_connector.py:30]: We can't really do that since: 
```
>>> not bool(None)
True
```

But we can do 
```suggestion
    if elem == {} or elem == []:
```

Sanix-Darker: Fair enough 👌🏽

aklom: Finalement, je pense qu'il ne faut pas rajouter le check sur le `[]` parce qu'on ne veut pas virer les steps où on a par exemple: `{'some_column': {'$in': []}`. ça va donner un résultat qui est faux.

PR #937, opened 2023-02-20
feat: mysql sslmode allow optional fields [TCTC-5306]

Sanix-Darker:
## WHAT

Just allow optional fields for the ssl mode on mysql

### DIFF
--- a/tests/mysql/test_mysql.py
+++ b/tests/mysql/test_mysql.py
@@ -483,17 +483,6 @@ def test_ssl_parameters_verify_identity(
     assert kwargs['ssl_verify_identity'] is True
 
 
-@pytest.mark.parametrize('missing_param', ('ssl_ca', 'ssl_cert', 'ssl_key'))
-def test_ssl_parameters_missing_param(
-    mysql_connector_with_ssl: MySQLConnector, mocker: MockerFixture, missing_param: str
-):
-    connect_mock = mocker.patch('pymysql.connect')
-    setattr(mysql_connector_with_ssl, missing_param, None)
-    with pytest.raises(AssertionError):
-        mysql_connector_with_ssl._connect()
-    assert connect_mock.call_count == 0
-
-
 def test_sanitize_ssl_params_invalid_pem(mysql_connector_with_ssl: MySQLConnector):
     mysql_connector_with_ssl.ssl_key = SecretStr(
         mysql_connector_with_ssl.ssl_key.get_secret_value()[:40]
--- a/toucan_connectors/mysql/mysql_connector.py
+++ b/toucan_connectors/mysql/mysql_connector.py
@@ -126,21 +126,25 @@ class MySQLConnector(ToucanConnector, DiscoverableConnector, VersionableEngineCo
         None,
         description='The CA certificate content in PEM format to use to connect to the MySQL '
         'server. Equivalent of the --ssl-ca option of the MySQL client',
+        required=False,
     )
     ssl_cert: SecretStr = Field(
         None,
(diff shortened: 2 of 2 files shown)


toucan_connectors/mysql/mysql_connector.py
```diff
@@ -126,21 +126,25 @@ class MySQLConnector(ToucanConnector, DiscoverableConnector, VersionableEngineCo
         None,
         description='The CA certificate content in PEM format to use to connect to the MySQL '
         'server. Equivalent of the --ssl-ca option of the MySQL client',
+        required=False,
```
lukapeschke, line 129: These are not needed, we already have `None` as the default value

toucan_connectors/mysql/mysql_connector.py
```diff
@@ -151,16 +155,12 @@ def _sanitize_ssl_params(self) -> dict[str, Any]:
...
-                assert (
-                    opt := getattr(self, ssl_opt)
-                ) is not None, (
-                    f'{ssl_opt} must be specified if ssl_mode is VERIFY_CA or VERIFY_IDENTITY'
-                )
                 try:
-                    params[ssl_opt] = sanitize_spaces_pem(opt.get_secret_value())
-
+                    if (opt := getattr(self, ssl_opt)) is not None:
+                        params[ssl_opt] = sanitize_spaces_pem(opt.get_secret_value())
```
lukapeschke, line 160: I'd systematically put the parameter in the dict for consistency.

Also, does it make sense to have only the key or the cert ? Shouldn't we check that if one is specified, the other one should be specified as well ?

toucan_connectors/mysql/mysql_connector.py
```diff
@@ -237,17 +237,20 @@ def _connect(
...
-                ssl_key.seek(0)
+                if 'ssl_ca' in ssl_params:
+                    ssl_ca.write(ssl_params['ssl_ca'].encode())
+                    ssl_ca.seek(0)
+                if 'ssl_cert' in ssl_params:
+                    ssl_cert.write(ssl_params['ssl_cert'].encode())
+                    ssl_cert.seek(0)
+                if 'ssl_key' in ssl_params:
+                    ssl_key.write(ssl_params['ssl_key'].encode())
+                    ssl_key.seek(0)
```
lukapeschke, line 248: You're still creating the temp files here. Maybe this can be refactored to a loop ?

toucan_connectors/mysql/mysql_connector.py
```diff
@@ -237,17 +237,20 @@ def _connect(
...
 
                 connection_params |= {
```
lukapeschke, line 250: I wonder wether `ssl_verify_cert` shouldn't be set to `True` as well if we  have no client key/cert

tests/mysql/test_mysql.py
```diff
@@ -483,17 +483,6 @@ def test_ssl_parameters_verify_identity(
...
-def test_ssl_parameters_missing_param(
-    mysql_connector_with_ssl: MySQLConnector, mocker: MockerFixture, missing_param: str
-):
-    connect_mock = mocker.patch('pymysql.connect')
-    setattr(mysql_connector_with_ssl, missing_param, None)
-    with pytest.raises(AssertionError):
-        mysql_connector_with_ssl._connect()
-    assert connect_mock.call_count == 0
-
-
```
lukapeschke, line 496: we still need a test checking the client key/cert pair integrity

lukapeschke (commented):
A few inline comments. Also, the target branch should be `release/v3.23`

PR #943, opened 2023-02-21
feat: implemented a hubspot connector working with private apps TCTC-5345

lukapeschke:
## Checklist


toucan_connectors/hubspot_private_app/hubspot_connector.py
```diff
@@ -0,0 +1,150 @@
...
+        ...
+
+
+class _Api(Protocol):  # pragma: no cover
+    @property
+    def basic_api(self) -> _BasicApi:
+        ...
+
+    def get_all(self) -> list[_HubSpotObject]:
+        ...
```
jGundermann, line 76: If i understand correctly this are the object from the hubspot api ?

Would this make sense to move this in another file and create utility function to manipulate them ?

lukapeschke: There are two things here:
* The pydantic objects are indeed used to parse the hubspot API response
* The `Protocol` objects are pure typing: Since the hubspot client does not provide type stubs, these are type definitions for it 

They _could_ be moved to a submodule, but since this file is so short, I figured that it would be more comfortable for reading to have everything in the same place. No strong opinion on this though

toucan_connectors/hubspot_private_app/hubspot_connector.py
```diff
@@ -0,0 +1,148 @@
...
+            results = []
+            result_iterator = self._result_iterator(
+                api, max_results=offset + (limit or 0), limit=limit
+            )
+            try:
+                for _ in range(offset):
+                    next(result_iterator)
+                results = list(result_iterator)
+            except StopIteration:
+                pass
```
Sanix-Darker, line 142: Is it not better like this ? 🤔 
```suggestion
           except StopIteration:
                pass
           results = list(result_iterator)
```

If there is an exception, the results will not be set right ?

tests/hubspot_private_app/test_hubspot.py
```diff
@@ -0,0 +1,214 @@
...
+
+from toucan_connectors.hubspot_private_app.hubspot_connector import (
+    HubspotConnector,
+    HubspotDataSource,
+)
+
+
+class _DictableDict(dict):
+    def to_dict(self) -> dict[str, Any]:
+        return self
```
Sanix-Darker, line 19: 😬 ➿

toucan_connectors/hubspot_private_app/hubspot_connector.py
```diff
@@ -103,6 +103,9 @@ def _result_iterator(
     ) -> Generator[_HubSpotResult, None, None]:
         after = None
         count = 0
+        # HubSpot returns a 400 HTTP error when trying to fetch 100 results
+        if limit is not None:
+            limit = min(limit, 100)
```
Sanix-Darker, line 108: then is it not better to have :
```suggestion
            limit = min(limit, 99)
```

Since 100results -> cause a 400 error...

just by curiosity... what's the message in the response for that 400 error... please ? 🤔  Is it a know issue or... ?

toucan_connectors/hubspot_private_app/hubspot_connector.py
```diff
@@ -90,7 +90,7 @@ def _fetch_all(self, api: _Api) -> list[_HubSpotResult]:
         results = api.get_all()
         return [_HubSpotResult(**elem.to_dict()) for elem in results]
 
-    def _api_for_dataset(self, object_type: HubspotDataset) -> _Api:
+    def _api_for_dataset(self, object_type: HubspotDataset) -> _Api:  # pragma: no cover
```
Sanix-Darker, line 93: the pragma: no cover is really needed here ? since the function start with the _ underscore... ? just asking... i sonar yelling here ?

tests/hubspot_private_app/test_hubspot.py
```diff
@@ -0,0 +1,214 @@
...
+            'hs_object_id': ['1', '51'],
+            'lastmodifieddate': ['2023-02-21T09:50:53.651Z', '2023-02-21T09:50:55.765Z'],
+            'lastname': ['Johnson (Sample Contact)', 'Halligan (Sample Contact)'],
+        }
+    )
+
+
+@pytest.fixture
+def hubspot_connector() -> HubspotConnector:
+    return HubspotConnector(name='hubspot', access_token='s3cr3t')
```
Sanix-Darker, line 128: is it all it need to instantiate it ? 🤔 ... what about the "data_source_model" or we don't care... it's a fixture

Sanix-Darker (commented):
Looks fine for me, i left some comments below !
is it available on an instance so that it can be tested... ?

toucan_connectors/hubspot_private_app/hubspot_connector.py
lukapeschke: If there's an exception, there's nothing left to read in the iterator, and `results` has already been assigned to an emtpy list on l.131

toucan_connectors/hubspot_private_app/hubspot_connector.py
lukapeschke: The message is "You cannot fetch more than 100 results"

toucan_connectors/hubspot_private_app/hubspot_connector.py
lukapeschke: Yup, we never execute it since it's mocked in the tests

toucan_connectors/hubspot_private_app/hubspot_connector.py
```diff
@@ -0,0 +1,150 @@
...
+    def _retrieve_data(self, data_source: HubspotDataSource) -> pd.DataFrame:
+        api = self._api_for_dataset(data_source.dataset)
+        return pd.DataFrame(r.to_dict() for r in self._fetch_all(api))
+
+    def _result_iterator(
+        self, api: _Api, max_results: int | None, limit: int | None
+    ) -> Generator[_HubSpotResult, None, None]:
+        after = None
+        count = 0
+        # HubSpot returns a 400 HTTP error when trying to fetch 100 results
```
lukapeschke, line 106:
```suggestion
        # HubSpot returns a 400 HTTP error when trying to fetch more than 100 results
```

toucan_connectors/hubspot_private_app/hubspot_connector.py
lukapeschke: adapted the comment to be clearer

Sanix-Darker (approved):

PR #948, opened 2023-02-22
feat: setUp dependabot automerge on patches

Sanix-Darker:
## WHAT

SetUp dependabot automerge on patches, since it's mostly fixes and small updates, 
we don't really want to have to check those everytime

## HOW

I added a new simple workflow that doesn't need a github app to process with an if condition on the semver bump.


Sanix-Darker:
No needed anymore for now.

PR #949, opened 2023-02-22
chore: v3.23.10

Sanix-Darker:
## WHAT

Bump from 3.23.9 to 3.23.10


lukapeschke:
Could we please wait for #943 to be merged and backported before making a release please ?

PR #982, opened 2023-03-07
chore: Loosen the upper python constraint

lukapeschke:
Note that this does not add official support for python 3.11 yet. For that, we must wait for awswrangler 3.0.0 to be released


pyproject.toml
```diff
@@ -21,7 +21,7 @@ readme = "README.md"
 # <3.11 required by awswrangler < 3.0
 python = ">=3.10,<3.12"
 Authlib = "^1.0.1"
-aiohttp = "^3.8.1"
+aiohttp = { version = "^3.8.1", python = "<3.11" }
```
Sanix-Darker, line 24: Can you also add a FIXME comment here ?

pyproject.toml
```diff
@@ -63,7 +63,7 @@ simplejson = {version = "^3.17.6", optional = true}
 pyhdb = {version = ">=0.3.4,<1.0", optional = true}
 zeep = {version = "^4.1.0", optional = true}
 snowflake-connector-python = {version = ">=2.7.12,<4.0.0", optional = true}
-pyarrow = {version = "<9", optional = true}
+pyarrow = {version = "<11", optional = true}
```
Sanix-Darker, line 66: Can you add also a fixme comment here ?

Sanix-Darker (commented):
Apart from comments... every thing looks fine for me... are we sure this is running inside a guru ? before merging this PR !

Sanix-Darker (approved):

PR #1057, opened 2023-04-27
fix(deps): Add missing aiohttp dependency 

lukapeschke:


Sanix-Darker (commented):
How the CI was green with missing deps...?
Or it was shadow installed by another deps ?

lukapeschke:
Yup it was an indirect depends of the extras

Sanix-Darker (approved):
Thanks for this 🦾

PR #1126, opened 2023-06-16
feat: google big query database selection much faster [TCTC-6113]

Sanix-Darker:
## WHAT

In case of a big project structure, we were struggling to fetch datasets using the
`_get_project_structure_fast` method in some huge cases, even tho it was implemented
too to speed up that process.
So, to solve that problem of latency, We tested numerous solutions such as *paginator_iter* from gbq lib,
update the fetching page per page and finaly, this is how we optimized it in 2 steps :
- For the method `_get_project_structure_fast`, if the project structure is expected to be large, we might want to paginate the results instead of returning a single DataFrame to avoid memory issues. that's why we added a query_job with the new parameter in the config : `_PAGE_SIZE`, this will help us paginate results regarding that parameter and `_MAXIMUM_RESULTS_FETCHED`, to impose ourself a limit of max results we can get from a single query
- We built up a generator like method to yield those results in a seperate function `_fetch_query_results`
- Add threading when fetching result by updating the `startIndex` parameter for each iteration.

## EDIT

- We return the schema from the list of databases to have the graphical interface(but this need a check on the frontend)

## EDIT 2

- We finally decided to retrieve the list of all schemas db first to let the user check what he/she wants when getting the model (project structure with the list of tables).

### DIFF
--- a/toucan_connectors/google_big_query/google_big_query_connector.py
+++ b/toucan_connectors/google_big_query/google_big_query_connector.py
@@ -48,13 +48,13 @@ class GoogleBigQueryDataSource(ToucanDataSource):
         **{'ui.hidden': True},
     )
     language: str = Field('sql', **{'ui.hidden': True})
-    database: str = Field(None, description='The name of the database you want to query.')
+    db_schema: str = Field(None, description='The name of the db_schema you want to query.')
 
     @classmethod
     def get_form(cls, connector: 'GoogleBigQueryConnector', current_config: dict[str, Any]):
         return create_model(
             'FormSchema',
-            database=strlist_to_enum('database', connector.available_dbs),
+            db_schema=strlist_to_enum('db_schema', connector.available_dbs),
             __base__=cls,
         ).schema()
 
--- a/toucan_connectors/google_big_query/google_big_query_connector.py
+++ b/toucan_connectors/google_big_query/google_big_query_connector.py
@@ -319,12 +319,7 @@ WHERE
         client = bigquery.Client(location=None, credentials=credentials)
         db_list = [ds.dataset_id for ds in client.list_datasets()]
 
-        print('\n' * 10)
-        print('-' * 100)
-        print(f'>> {db_list=}')
         db_as_series = pd.Series(db_list).values
-        print('-' * 100)
-        print(f'>> {db_as_series=}')
 
         return db_as_series
 
--- a/toucan_connectors/google_big_query/google_big_query_connector.py
+++ b/toucan_connectors/google_big_query/google_big_query_connector.py
@@ -52,11 +52,14 @@ class GoogleBigQueryDataSource(ToucanDataSource):
 
     @classmethod
     def get_form(cls, connector: 'GoogleBigQueryConnector', current_config: dict[str, Any]):
-        return create_model(
+        schema = create_model(
             'FormSchema',
             db_schema=strlist_to_enum('db_schema', connector.available_dbs),
             __base__=cls,
         ).schema()
+        schema['properties']['database']['default'] = connector.credentials.project_id
+
+        return schema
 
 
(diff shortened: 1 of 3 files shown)


toucan_connectors/google_big_query/google_big_query_connector.py
```diff
@@ -226,21 +230,45 @@ def _format_columns(x: str):
...
         )
 
+    def _fetch_query_results(self, query_job: QueryJob) -> Generator[Any, Any, Any]:
+        """Fetches query results in a paginated manner using a generator."""
+        row_iterator = query_job.result(page_size=_PAGE_SIZE, max_results=_MAXIMUM_RESULTS_FETCHED)
+
+        while True:
+            rows = next(row_iterator.pages, None)
+            if rows is None:
+                break
```
lukapeschke, line 269:
```suggestion
        while rows := next(row_iterator.pages, None):
```

lukapeschke (commented):
Thanks for the improvements, looks good :muscle: The only thing that bothers me is that I couldn't find documentation about `IS_INDEXABLE`, do you have a link to the docs ? We wouldn't want to hide columns by accident

fspot (approved):
LGTM :+1:

toucan_connectors/google_big_query/google_big_query_connector.py
```diff
@@ -226,21 +232,109 @@ def _format_columns(x: str):
...
+    def _merge_query_results(self, query: str, client: bigquery.Client, num_threads: int = 6):
+        """Execute query with multiple threads, incrementing startIndex, and merge results."""
+        lock = threading.Lock()
+
+        results, threads = [], []
+
+        # Create the query job with the initial startIndex
+        # Get the total rows count
+        query_job = client.query(query)
+        total_rows = getattr(query_job, 'total_rows', _MAXIMUM_RESULTS_FETCHED)
```
lukapeschke, line 299: Is this always present ? If not, we might end up with a few "useless" threads, if we go with 2000 / N

toucan_connectors/google_big_query/google_big_query_connector.py
```diff
@@ -226,21 +232,109 @@ def _format_columns(x: str):
...
+
+        for thread in threads:
+            thread.start()
+
+        # Wait for all threads to finish
+        for thread in threads:
+            thread.join()
+
+        return pd.DataFrame(results)
+
```
lukapeschke, line 320: I'm not sure about multi-threading: It does not necessarily play well when used in an API.  Maybe enabling mutlithreading should be an option of the connector ? Along with the max number of threads

Also, If you want to go with multithreading, the the best way to do it would probably be with a `ThreadPoolExecutor`. It'd avoid the mutex and allows to re-use existing threads

toucan_connectors/google_big_query/google_big_query_connector.py
```diff
@@ -226,21 +232,109 @@ def _format_columns(x: str):
...
+                page_size=_PAGE_SIZE,
+                max_results=_MAXIMUM_RESULTS_FETCHED,
+            )
+
+            while rows := next(row_iterator.pages, None):
+                yield rows.to_dataframe()
+        except GoogleExceptionNotfound as excp:
+            # we ignore when we don't found a dataset on a locationk
+            # google.api_core.exceptions.NotFound:
+            # 404 Not found: Dataset data-finance-staging:dbt_yanni_iyeze_dbt_test__audit was not found in location EU
```
lukapeschke, line 335: Can you please anonymize the dataset name ?

Also, I beleive that the exception must not be caught here: For clients with datasets in several regions, only those from the first encountered region will appear (and `_get_project_structure_slow` will never be called)

lukapeschke (dismissed):

fspot (approved):
Ok pour moi

fspot (approved):

PR #1273, opened 2023-09-11
feat: add s3 connector TCTC-6566

andreamouraud:
## Change Summary

Add a S3 connector using the AWS Security Token Service API 

## Set up 
The following properties need to be injected into the connector by the backend:
- `sts_access_key_id`
- `sts_secret_access_key`
- `workspace_id`

The `external_id` field is read only and represents the injected `workspace_id`, it is used to create the corresponding AWS role

## Retrieving data

The file specified in the `S3DataSource` is loaded from the S3 bucket using the injected credentials, https://github.com/ToucanToco/peakina is then used to load the DataFrame from the file

## Checklist


toucan_connectors/s3/s3_connector.py
```diff
@@ -0,0 +1,164 @@
...
+    def _sts_assumed_role(self) -> dict:
+        sts_client = boto3.client(
+            'sts',
+            aws_access_key_id=self.sts_access_key_id,
+            aws_secret_access_key=self.sts_secret_access_key,
+        )
+        return sts_client.assume_role(
+            RoleArn=self.role_arn,
+            RoleSessionName='toucantoco',
+            ExternalId=self.workspace_id,
```
Sanix-Darker, line 152:
```suggestion
            ExternalId=self.external_id,
```

toucan_connectors/s3/s3_connector.py
```diff
@@ -0,0 +1,164 @@
...
+        )
+
+    def _get_assumed_sts_role(self) -> dict:
+        with suppress(Exception):
```
Sanix-Darker, line 156:
```suggestion
        with suppress(KeyError):
```
to me more explicit ?

toucan_connectors/s3/s3_connector.py
```diff
@@ -0,0 +1,164 @@
...
+        fetcher_kwargs = {
+            **fetcher_kwargs,
+            'client_kwargs': {**client_kwargs, 'session_token': session_token},
+        }
+
+        session = boto3.Session(
+            aws_access_key_id=access_key,
+            aws_secret_access_key=access_secret,
+            aws_session_token=session_token,
+        )
```
Sanix-Darker, line 113: should we not add this in a `@cached_property` as well too ? (with lines 89-95)

toucan_connectors/s3/s3_connector.py
andreamouraud: The `external_id` property is only used as a read-only field for the UI, what's actually used is the workspace id, in theory the two of them always resolve to the same value

Sanix-Darker: after checking tests it was much clear for me, thanks for clarifications

toucan_connectors/s3/s3_connector.py
andreamouraud: we don't want to be specific here, whatever exception is raised we want to suppress it
i don't actually think anything else could be raised for now but it seems more explicit to me this way, wyt?

toucan_connectors/s3/s3_connector.py
andreamouraud: i thought of caching the whole s3 client, but as i don't think its actually creating any http calls in the background it shouldn't be such a big deal and it keeps the code easy to read
might want to double check it doesn't doesn't have any http call, if it does it makes sense to cache it yeah

toucan_connectors/s3/s3_connector.py
Sanix-Darker: let's keep it simple then with only "Exception" 👌🏽

toucan_connectors/s3/s3_connector.py
andreamouraud: @Sanix-Darker just verified it:
```python
session = boto3.Session(
    aws_access_key_id=access_key,
    aws_secret_access_key=access_secret,
    aws_session_token=session_token,
)
s3 = session.client('s3')
```
- no http request is being made
-  simple class initialisation and set up 
- longest operations are botocore internals JSON files being loaded
-  both session + client init takes about 90ms on my machine (including debug logging)

No need to make the code more complicated and add caching IMO

Sanix-Darker: looks great then ! 👌🏽

andreamouraud:
@Sanix-Darker great idea, will do in the bump 👍

PR #1293, opened 2023-09-21
fix(gbq): Validation error on mandatory GoogleCreds missing + add descriptive Exception message on missing datas [TCTC-6756]

Sanix-Darker:
## WHAT

- Explicit Exception information when data are not available when an ValueError ( on empty data in the dataframe when concatenate)
- Fix a ValidationError introduced in the previous bump (All GoogleCredentials should no be "mandatory"/"required" for the GoogleBigQuery connector)
- Add a fallback on GoogleCredentials login, so that if the jwt-token is not valid anymore, we can still connect with GoogleCreds if they are available

### DIFF
--- a/toucan_connectors/google_big_query/google_big_query_connector.py
+++ b/toucan_connectors/google_big_query/google_big_query_connector.py
@@ -39,6 +39,10 @@ class InvalidJWTToken(GoogleUnauthorized):
     """When the jwt-token is no longer valid (usualy from google as 401)"""
 
 
+class GoogleClientCreationError(Exception):
+    """When it's not possible to create a bigquery.Client with givent fields"""
+
+
 class NoDataFoundException(Exception):
     """When there is no data to Concatenate and send back to the user"""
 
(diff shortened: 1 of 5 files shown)


toucan_connectors/google_big_query/google_big_query_connector.py
```diff
@@ -112,6 +120,35 @@ def _define_query_param(name: str, value: Any) -> BigQueryParam:
...
+            'client_email',
+            'client_id',
+            'auth_uri',
+            'auth_provider_x509_cert_url',
+            'client_x509_cert_url',
+        ]
+
+        # Iterate over optional fields and set them to None if missing
+        for field_name in optional_fields:
+            if not hasattr(value, field_name) and not value.get(field_name):
```
andreamouraud, line 138:
```suggestion
            if not hasattr(value, field_name):
```

toucan_connectors/google_big_query/google_big_query_connector.py
```diff
@@ -278,20 +321,35 @@ def _bigquery_variable_transformer(variable: str):
...
+                # We fallback the login on GoogleCredentials creation of the client here ?
+                # or we fallback on default google-credentials
+                client = self._bigquery_client_with_google_creds()
         else:
             # or we fallback on default google-credentials
-            credentials = GoogleBigQueryConnector._get_google_credentials(
-                self.credentials, self.scopes
-            )
-            client = GoogleBigQueryConnector._connect(credentials)
+            client = self._bigquery_client_with_google_creds()
```
andreamouraud, line 352:
```suggestion
        # We fallback the login on GoogleCredentials creation of the client here ?
        # or we fallback on default google-credentials
        client = self._bigquery_client_with_google_creds()
```

toucan_connectors/google_big_query/google_big_query_connector.py
```diff
@@ -112,6 +120,35 @@ def _define_query_param(name: str, value: Any) -> BigQueryParam:
...
+                # for all links
+                if field_name in [
+                    'client_x509_cert_url',
+                    'auth_provider_x509_cert_url',
+                    'auth_uri',
+                ]:
+                    value[field_name] = 'https://valid-scheme.com'
+                else:
+                    # for other values
+                    value[field_name] = '__not_set__'
```
andreamouraud, line 148: is there no other way of doing that? seems a bit odd

andreamouraud (changes requested):

toucan_connectors/google_big_query/google_big_query_connector.py
Sanix-Darker: And you're right, after discussing with Thomas, we all agreed we should have a field "project-id" for the jwt,  so I created a second model called JWTCredentials and remove all those "WEIRDS" validators we were having

Sanix-Darker: Please see below a demo : 
test
let me know what you think

Sanix-Darker: Either we check for the "GoogleCredentials authentification" either we check for the "JWTCredentials authentication" and then fill all those required values.

andreamouraud (approved):

PR #1333, opened 2023-10-20
ci: v4.9.3

davinov:


CHANGELOG.md
```diff
@@ -1,6 +1,6 @@
 # Changelog (Pypi package)
 
-## Unreleased
+### [4.9.3] 2023-10-20
```
davinov, line 3: Noticed that the whole file is incorrectly formatted (version title should be h2 and fixed, changed, etc. should be h3) but that's not the scope of this PR !

lukapeschke (approved):

PR #1406, opened 2023-12-15
chore(deps): replace black+isort+flake8 with ruff

lukapeschke:


davinov (approved):
Let's merge it despite Sonar alert on coverage: this is not new code, but lines have sometimes changed dur to reformatting.

PR #1469, opened 2024-01-23
fix(elasticsearch): disable verification

WilliamGorge:
## Change Summary
Disable the internal verification of the elasticsearch client that checks if the root `/` of the provided URL returns a a compatible payload.

This internal verification, for some reason, does not use the basic auth provided by http_auth. So this check fails when the client configuration ask for authentication for the `/` route.

https://github.com/elastic/elasticsearch-py/blob/v7.17.6/elasticsearch/transport.py#L532
https://github.com/elastic/elasticsearch-py/blob/v7.17.6/elasticsearch/transport.py#L216

https://toucantoco.slack.com/archives/C823R095H/p1705597838374409

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ### Changed
 
 - Google Big Query: A simple status check that validates the private key's format has been implemented
+- Elasticsearch: Host verification has been disabled to tolerate strict network configurations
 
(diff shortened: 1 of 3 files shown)


lukapeschke (approved):

WilliamGorge (commented):
LGTM: merci pour le com

PR #1475, opened 2024-01-23
feat(google_big_query): actual connection check

davinov:
With a very simple query, that works even without any table. When testing, I measured that the status check took less than 1s.


lukapeschke (approved):
:1st_place_medal: niiiiiiice

PR #1514, opened 2024-02-09
feat(mongo_connector): Make pool size configurable

lukapeschke:
Work items:

- Added a new `max_pool_size` parameter to MongoConnector, which defaults to 1 (PyMongo's default pool size is 100).
- Ensure mongo clients are closed after usage. This may seem sub-optimal, but a typical connector usage is instanciation -> call a single method (retrieve data or get_status) -> close the client, so this seems reasonable
- Enable mypy checks on the Mongo connector, since it's by far the one we use the most

relates to TCTC-7814

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### Changed
+
+- Mongo: maximal connection pool size is now configurable via the `max_pool_size` parameter. It defaults to 1
+
 ## [5.2.0] 2024-02-08
 
 ### Changed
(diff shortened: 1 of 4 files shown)


lukapeschke:
# Connection beahvour on master
https://github.com/ToucanToco/toucan-connectors/assets/17085536/e06e7cb5-fb6b-493b-8f5c-21714723a2c5

# Connection behaviour on this branch

https://github.com/ToucanToco/toucan-connectors/assets/17085536/194d0968-b935-4d8d-875e-c20ebd203ac8

PR #1580, opened 2024-03-14
fix: elasticsearch form schema [TCTC-8032]

fspot:
Capturevido2024-03-14152813-ezgif com-video-to-gif-converter


lukapeschke (approved):
Thx! does this need a backport ?

PR #1666, opened 2024-06-07
fix(oracle_sql): pass parameters to pandas read_sql as numeric parameters

julien-pinchelimouroux:
## Change Summary

 - OracleSQL: pass parameters as numeric to `read_sql`

### DIFF
--- a/toucan_connectors/common.py
+++ b/toucan_connectors/common.py
@@ -340,6 +340,45 @@ def convert_to_qmark_paramstyle(query_string: str, params_values: dict) -> tuple
     return re.sub(RE_NAMED_PARAM, "?", query_string), flattened_values
 
 
+def convert_to_numbered_vars_paramstyle(query_string: str, params_values: dict) -> tuple[str, tuple[Any]]:
+    """Takes a query in pyformat paramstyle and transforms it in numbered paramstyle
+       by replacing placeholders by :n and returning values in right order
+    ex :
+        ('select * from test where id > %(id_nb)s and price > %(price)s;', {"id_nb":1, "price":10}
+    returns:
+        ('select * from test where id > :1 and price > :2;', (1, 10))"""
+    extracted_params = re.findall(RE_NAMED_PARAM, query_string)
+    qparams = [get_param_name(m) for m in extracted_params]
+    ordered_values = [params_values.get(p) for p in qparams]
+
+    variable_idx = 1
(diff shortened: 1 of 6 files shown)


toucan_connectors/oracle_sql/oracle_sql_connector.py
```diff
@@ -80,7 +80,14 @@ def _retrieve_data(self, data_source: OracleSQLDataSource) -> pd.DataFrame:
...
         query = data_source.query[:-1] if data_source.query.endswith(";") else data_source.query
-        df = pandas_read_sql(query, con=connection)
+        query_params = data_source.parameters or {}
+        df = pandas_read_sql(
+            query,
+            con=connection,
+            params=query_params,
+            render_user=True,
+            convert_to_numbered=True,
+        )
```
lukapeschke, line 90: Should you also specify `convert_to_printf=False` here ?

toucan_connectors/common.py
```diff
@@ -340,6 +340,45 @@ def convert_to_qmark_paramstyle(query_string: str, params_values: dict) -> tuple
     return re.sub(RE_NAMED_PARAM, "?", query_string), flattened_values
 
 
+def convert_to_numbered_vars_paramstyle(query_string: str, params_values: dict) -> tuple[str, tuple[Any]]:
```
lukapeschke, line 343: It's called the `numeric` param style: https://peps.python.org/pep-0249/#paramstyle

Also, could you please add parametrized unit tests for this function ? If the oracle connectors is discontinued at some point, we won't have any tests left

lukapeschke (changes requested):

toucan_connectors/common.py
```diff
@@ -340,6 +340,45 @@ def convert_to_qmark_paramstyle(query_string: str, params_values: dict) -> tuple
     return re.sub(RE_NAMED_PARAM, "?", query_string), flattened_values
 
 
+def convert_to_numeric_paramstyle(query_string: str, params_values: dict) -> tuple[str, tuple[Any]]:
+    """Takes a query in pyformat paramstyle and transforms it in numbered paramstyle
```
fspot, line 344:
```suggestion
    """Takes a query in pyformat paramstyle and transforms it in numeric paramstyle
```

fspot (approved):

lukapeschke (approved):

PR #1668, opened 2024-06-12
chore: v6.2.0

julien-pinchelimouroux:


lukapeschke (changes requested):
sonar properties need to be updated as well

lukapeschke (commented):
Also it should be 6.2.0, since a feature was introduced

lukapeschke (approved):

PR #1671, opened 2024-06-13
chore: v3.23.30

julien-pinchelimouroux:


lukapeschke (changes requested):
This branch contains extremely cursed and non-semver versioning, so we stick to `3.23.xxx`. Changes are correct but version should be 3.23.30

PR #1751, opened 2024-08-29
feat: add token_header_name option to http connectors (TCTC-8945)

julien-pinchelimouroux:
## Change Summary

- Adds `token_header_name` option to `CustomServerToken` authentication method.


julien-pinchelimouroux:
Add option to connector's documentation
Specify that CustomTokenServer endpoint needs to return a json object with auth token inside

PR #1764, opened 2024-09-10
fix(mysql): get_model now supports extra kwargs so it doesnt crash [TCTC-9223]

fspot:
Laputa now calls get_model with some extra parameters, but these parameters were not added for 3 sql connector's `get_model` method : 
- mysql
- redshift
- snowflake oauth2

When get_model fails, the screen supposed to list db tree structure loads forever (cf. jira card https://toucantoco.atlassian.net/browse/TCTC-9223)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### Fix
+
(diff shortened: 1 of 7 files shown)


lukapeschke (approved):
My bad I was 100 sure that I did the update everywhere :facepalm: 

Could we maybe enable type checking on these files if it isn't too much work ?

PR #1772, opened 2024-09-13
feat: when testing connection, set timeout to 10s when checking if port is opened [TCTC-7385]

fspot:
https://toucantoco.atlassian.net/browse/TCTC-7385


julien-pinchelimouroux (approved):
:ok_hand:

PR #1784, opened 2024-10-02
fix(biquery): enforce int and float dtypes [TCTC-9419]

lukapeschke:
Also introduce integration tests for GBQ


julien-pinchelimouroux (approved):
Merci ! :pray:

PR #1794, opened 2024-10-14
feat: add pagination configs for HTTP API Connector (TCTC-9227)

julien-pinchelimouroux:
## Proposal pagination v0: 

Add a `http_Pagination` attribute to `HttpAPIConnector` which works like the `auth` configuration:
 - the user can activate pagination in its interface and then choose the kind of pagination he wants.
 - each pagination type must have its default configuration, which will be displayed to the user once its selected
 
 That `http_pagination` is transformed to a `PaginationConfig`. 
 All PaginationConfigs works the same : 
  - `apply_pagination_to_data_source` : it transform the data_source in order to add all pagination arguments needed to the API request
  - `get_next_pagination_config`: it computes the next pagination_config for the next API call based on the last API response
  - [optional] `get_pagination_info_filter`: it returns the JQ filter that can be used to extract pagination information from the previous API call
  
## Next steps: 
 - Implement more pagination types
 - Accept other HTTP method / parameter position (ex offset_limit in request headers)
 - Add  http_config form to the front end connector settings
 - Write pagination config documentation section
   
## Limits:
 - All HTTP calls will be performed one after the other, that can produce throttling errors (429). Maybe add an additional config for that in the http_config?
 - HTTP API must be able to return all results in less than 30s. We can only accept small datasets.
 
 ## Front example 
 
image

### DIFF
--- a/toucan_connectors/http_api/http_api_connector.py
+++ b/toucan_connectors/http_api/http_api_connector.py
@@ -8,7 +8,7 @@ from pydantic import AnyHttpUrl, BaseModel, Field, FilePath
 from requests.exceptions import HTTPError
 
 from toucan_connectors.http_api.http_api_data_souce import HttpAPIDataSource
-from toucan_connectors.http_api.pagination_configs import PaginationConfig
+from toucan_connectors.http_api.pagination_configs import HttpPagination, PaginationConfig
 
 try:
     import pandas as pd
@@ -70,7 +70,7 @@ class HttpAPIConnector(ToucanConnector, data_source_model=HttpAPIDataSource):
         None,
         description="You can provide a custom template that will be used for every HTTP request",
     )
-    pagination_config: PaginationConfig = Field(PaginationConfig(), title="Pagination configuration")
+    http_pagination: HttpPagination = Field(HttpPagination(), title="Pagination configuration")
 
     def do_request(self, query, session):
         """
@@ -155,7 +155,9 @@ class HttpAPIConnector(ToucanConnector, data_source_model=HttpAPIDataSource):
         try:
             results = pd.DataFrame(
                 self.perform_requests(
-                    data_source=data_source, session=session, pagination_config=self.pagination_config
+                    data_source=data_source,
+                    session=session,
+                    pagination_config=self.http_pagination.get_pagination_config(),
                 )
             )
         except HTTPError as exc:
--- a/toucan_connectors/http_api/http_api_connector.py
+++ b/toucan_connectors/http_api/http_api_connector.py
@@ -26,6 +26,7 @@ except ImportError as exc:  # pragma: no cover
 
 from toucan_connectors.auth import Auth
 from toucan_connectors.common import (
+    HttpError,
     nosql_apply_parameters_to_query,
     transform_with_jq,
 )
@@ -37,6 +38,10 @@ TOO_MANY_REQUESTS = 429
 
 class HttpAPIConnectorError(Exception):
     """Raised when an error occurs while fetching data from an HTTP API"""
+
+    def __init__(self, message: str, original_exc: HttpError) -> None:
+        super().__init__(message)
+        self.original_exc = original_exc
 
 
 class ResponseType(str, Enum):
@@ -182,8 +187,9 @@ class HttpAPIConnector(ToucanConnector, data_source_model=HttpAPIDataSource):
         except HTTPError as exc:
             if exc.response.status_code == TOO_MANY_REQUESTS:
                 raise HttpAPIConnectorError(
-                    "Failed to retrieve data: the connector tried to perform too many requests."
-                    " Please check your API call limitations."
+                    message="Failed to retrieve data: the connector tried to perform too many requests."
+                    " Please check your API call limitations.",
+                    original_exc=exc,
                 ) from exc
             else:
                 raise
--- a/toucan_connectors/http_api/http_api_connector.py
+++ b/toucan_connectors/http_api/http_api_connector.py
@@ -3,9 +3,12 @@ from enum import Enum
 from logging import getLogger
 from typing import Any, List
 from xml.etree.ElementTree import ParseError, fromstring, tostring
+from requests.exceptions import HTTPError
 
 from pydantic import AnyHttpUrl, BaseModel, Field, FilePath
-from pydantic.json_schema import DEFAULT_REF_TEMPLATE, GenerateJsonSchema, JsonSchemaMode
+
+from toucan_connectors.http_api.http_api_data_souce import HttpAPIDataSource
+from toucan_connectors.http_api.pagination_configs import PaginationConfig
 
 try:
     import pandas as pd
@@ -19,8 +22,6 @@ except ImportError as exc:  # pragma: no cover
 
 from toucan_connectors.auth import Auth
 from toucan_connectors.common import (
-    FilterSchema,
-    XpathSchema,
     nosql_apply_parameters_to_query,
     transform_with_jq,
 )
@@ -28,15 +29,12 @@ from toucan_connectors.toucan_connector import ToucanConnector, ToucanDataSource
 from toucan_connectors.utils.json_to_table import json_to_table
 
 
+TOO_MANY_REQUESTS = 429
+
+
 class ResponseType(str, Enum):
     json = "json"
     xml = "xml"
-
-
-class Method(str, Enum):
-    GET = "GET"
-    POST = "POST"
-    PUT = "PUT"
 
 
 class Template(BaseModel):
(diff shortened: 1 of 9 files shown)


toucan_connectors/http_api/http_api_connector.py
```diff
@@ -183,24 +115,63 @@ def do_request(self, query, session):
...
+            HttpAPIConnector.logger.error(f"Could not transform {raw_result} using {jq_filter}")
             raise
 
+        results = []
+        if next_pagination_config := pagination_config.get_next_pagination_config(
+            result=parsed_result, pagination_info=parsed_pagination_info
+        ):
+            results += self.perform_requests(
+                data_source=data_source, session=session, pagination_config=next_pagination_config
+            )
```
lukapeschke, line 144: Could we have something non-recursive here ? I'm afraid that for APis with very small page size (such as Hubspot), we'll quickly reach the stack limits :grimacing:

toucan_connectors/http_api/http_api_connector.py
```diff
@@ -183,24 +115,63 @@ def do_request(self, query, session):
...
+                    data_source=data_source,
+                    session=session,
+                    pagination_config=self.http_pagination.get_pagination_config(),
+                )
+            )
+        except HTTPError as exc:
+            if exc.response.status_code == TOO_MANY_REQUESTS:
+                raise HTTPError(
+                    "Fails to retrieve data: the connector tried to perform too many requests."
+                    " Please check your API call limitations."
```
lukapeschke, line 167:
```suggestion
                    "Failed to retrieve data: the connector tried to perform too many requests."
                    " Please check your API call limitations."
```

toucan_connectors/http_api/http_api_connector.py
```diff
@@ -183,24 +115,63 @@ def do_request(self, query, session):
...
+            results = pd.DataFrame(
+                self.perform_requests(
+                    data_source=data_source,
+                    session=session,
+                    pagination_config=self.http_pagination.get_pagination_config(),
+                )
+            )
+        except HTTPError as exc:
+            if exc.response.status_code == TOO_MANY_REQUESTS:
+                raise HTTPError(
```
lukapeschke, line 165: A custom exception should be raised here, as you're removing information from the original information, such as the status code: If some other code is catching `HTTPError` higher up, if might have unexpected behavior since some information it expects to find in the exception wouldn't be present

toucan_connectors/http_api/pagination_configs.py
```diff
@@ -0,0 +1,128 @@
+from enum import Enum
+from typing import Any, Optional
+
+from pydantic import BaseModel, Field
+
+from toucan_connectors.common import FilterSchema
+from toucan_connectors.http_api.http_api_data_souce import HttpAPIDataSource
+
+
+class PaginationConfig(BaseModel):
```
lukapeschke, line 10: Nit: could you make this an abstract class and create a `NoopPaginationConfig` to make it explicit that it will not paginate ?

toucan_connectors/http_api/pagination_configs.py
```diff
@@ -0,0 +1,128 @@
...
+    limit_name: str = "limit"
+    limit: int
+
+    def apply_pagination_to_data_source(self, data_source: HttpAPIDataSource) -> HttpAPIDataSource:
+        offset_limit_params = {self.offset_name: self.offset, self.limit_name: self.limit}
+        if data_source.params is None:
+            data_source.params = offset_limit_params
+        else:
+            data_source.params |= offset_limit_params
+        return data_source
```
lukapeschke, line 42: Implem looks good, but I'd prefer if it did not mutate the passed data source:

```suggestion
    def apply_pagination_to_data_source(self, data_source: HttpAPIDataSource) -> HttpAPIDataSource:
        offset_limit_params = {self.offset_name: self.offset, self.limit_name: self.limit}
        if data_source.params is None:
            params = offset_limit_params
        else:
            params |= offset_limit_params
        return data_source.model_copy(updates={"params": params})
```

toucan_connectors/http_api/pagination_configs.py
```diff
@@ -0,0 +1,128 @@
...
+    page_name: str = "page"
+    page: int = 0
+
+    def apply_pagination_to_data_source(self, data_source: HttpAPIDataSource) -> HttpAPIDataSource:
+        page_based_params = {self.page_name: self.page}
+        if data_source.params is None:
+            data_source.params = page_based_params
+        else:
+            data_source.params |= page_based_params
+        return data_source
```
lukapeschke, line 63: Same remark re: mutation

toucan_connectors/http_api/pagination_configs.py
```diff
@@ -0,0 +1,128 @@
...
+
+    def get_pagination_config(self) -> PaginationConfig:
+        auth_class = {
+            PaginationType.nothing: PaginationConfig,
+            PaginationType.offset_limit: OffsetLimitPaginationConfig,
+            PaginationType.cursor: CursorBasedPaginationConfig,
+            PaginationType.page: PageBasedPaginationConfig,
+        }[self.type]
+        pagination_config = auth_class(**self.kwargs)
+        return pagination_config
```
lukapeschke, line 128: Why not an union type ?

`HttpPaginationConfig: TypeAlias = PaginationConfig | OffsetLimitPaginationConfig ...`

toucan_connectors/http_api/http_api_connector.py
```diff
@@ -183,24 +115,63 @@ def do_request(self, query, session):
...
+        jq_filter = query["filter"]
+        raw_result = self.do_request(query, session)
+
         try:
-            return transform_with_jq(data, jq_filter)
+            parsed_result = transform_with_jq(raw_result, jq_filter)
+            parsed_pagination_info = None
+            if jq_pagination_filter := pagination_config.get_pagination_info_filter():
+                # transform_with_jq search for all occurrences, it will always return a list of values
+                parsed_pagination_info = transform_with_jq(raw_result, jq_pagination_filter)[0]
```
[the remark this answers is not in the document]
julien-pinchelimouroux: In fact, I don't think that try/except is useful
If we look at `test_get_df_with_cursor_pagination` test, even if the key is missing, `transform_with_jq` will return `[None]`, which will result as stopping the data fetching : `pagination_config.get_next_pagination_config()` will return None

toucan_connectors/http_api/http_api_connector.py
julien-pinchelimouroux: You're right, didnt thought to async loads with larger datasets !

toucan_connectors/http_api/pagination_configs.py
```diff
@@ -0,0 +1,128 @@
...
+        if data_source.params is None:
+            data_source.params = page_based_params
+        else:
+            data_source.params |= page_based_params
+        return data_source
+
+    def get_next_pagination_config(
+        self, result: Any, pagination_info: Any | None
+    ) -> Optional["PageBasedPaginationConfig"]:
+        if len(result) < 1:
```
davinov, line 68: Optional: maybe some API would return an error instead of an empty page? E.g. `404`

toucan_connectors/http_api/pagination_configs.py
davinov: That may also lead to a better generated form, with arguments being more explicit for each type.

davinov (commented):
Thanks for this good work :+1: 
I think it would be nice to have a document (PAT) somewhere to describe the pagination types, maybe with examples. That would help:
- us to find easily which cases are handled and which are not
- the product team to write the doc

davinov:
Also: there is a type of pagination, advised in Hypermedia APIs, where we get the next page URL: https://www.google.com/search?client=firefox-b-d&q=REST+hypermedia+pagination
I think it should be quite easy to implement, as `apply_pagination_to_data_source` should just replace all or a part of the URL (usually just keep the scheme and domain).
Do we think it would be worth adding it?

fspot:
Top :+1:

julien-pinchelimouroux:
P.A.T => https://toucantoco.atlassian.net/wiki/spaces/TTA/pages/4615897116/PAT+276+--+HTTP+Connector+v1+Pagination

davinov (commented):
Thanks! Could we add a log on each page, so we can follow the work done while it happens? (as it will probably take some time if there is many pages)

toucan_connectors/http_api/pagination_configs.py
```diff
@@ -0,0 +1,189 @@
...
+
+from pydantic import BaseModel, Field
+
+from toucan_connectors.common import FilterSchema
+from toucan_connectors.http_api.http_api_data_souce import HttpAPIDataSource
+
+_LOGGER = logging.getLogger(__name__)
+
+
+class PaginationConfig(BaseModel):
```
lukapeschke, line 13: Needs to inherit ABC as well to be considered abstract: 
* https://docs.python.org/3/library/abc.html 
* https://docs.pydantic.dev/latest/concepts/models/#abstract-base-classes

toucan_connectors/http_api/pagination_configs.py
```diff
@@ -0,0 +1,189 @@
...
+        """Computes next pagination config based on the parsed API responses"""
+        pass
+
+    def get_pagination_info_filter(self) -> str | None:
+        """Returns the JQ filter that must be applied to the raw API result to retrieve pagination info"""
+        pass
+
+    def get_error_status_whitelist(self) -> list[str] | None:
+        """Returns the list of the error statuses which means the end of data fetching, and so to ignore"""
+        pass
```
lukapeschke, line 34:
```suggestion
    @abstractmethod
    def apply_pagination_to_data_source(self, data_source: HttpAPIDataSource) -> HttpAPIDataSource:
        """Apply pagination to data source and return it"""

    @abstractmethod
    def get_next_pagination_config(self, result: Any, pagination_info: Any | None) -> Optional["PaginationConfig"]:
        """Computes next pagination config based on the parsed API responses"""

    @abstractmethod
    def get_pagination_info_filter(self) -> str | None:
        """Returns the JQ filter that must be applied to the raw API result to retrieve pagination info"""

    @abstractmethod
    def get_error_status_whitelist(self) -> list[str] | None:
        """Returns the list of the error statuses which means the end of data fetching, and so to ignore"""
```

toucan_connectors/http_api/http_api_connector.py
```diff
@@ -19,26 +26,24 @@
...
 )
 from toucan_connectors.toucan_connector import ToucanConnector, ToucanDataSource
 from toucan_connectors.utils.json_to_table import json_to_table
 
+TOO_MANY_REQUESTS = 429
+
+
+class HttpAPIConnectorError(Exception):
+    """Raised when an error occurs while fetching data from an HTTP API"""
+
```
lukapeschke, line 40: would be great if we had an `original_exc` attribute for later investigation:

```suggestion
class HttpAPIConnectorError(Exception):
    """Raised when an error occurs while fetching data from an HTTP API"""
    
    def __init__(self, original_exc: HttpError) -> None:
        super().__init__()
        self.original_exc = original_exc

```

lukapeschke (commented):
Much better, thanks a lot!

davinov (approved):

PR #1807, opened 2024-10-31
fix: merge api results after their transformations (TCTC-9634)

julien-pinchelimouroux:
## Why

API results where appended before any user's configured transformations (filters, flatten). We have to append resulted dataframes.
Moreover, APIs can return directly json objects, which created invalid results merging.
In addition, we have to add a data_length_filter to offset_limit pagination in case of the data result isn't relevant to compute the data length.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,11 @@
 
 ## Unreleased
 
+### Fix
+
+- HTTP API: Extract API results dataframes separately and merge them after.
+- HTTP API: Add `data_length_filter` offset pagination config field to determine which part of data must be used to compute the data length.
+
 ## [7.1.1] 2024-10-28
 
 ### Fix
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,8 +4,11 @@
 
 ### Fix
 
-- HTTP API: Extract API results dataframes separately and merge them after.
-- HTTP API: Add `data_length_filter` offset pagination config field to determine which part of data must be used to compute the data length.
+- HTTP API: API results are now correctly merged even if they need to be filtered or flattened. 
+
+### Added
+
+- HTTP API: Add `data_filter` offset pagination config field to determine which part of data must be used to compute the data length.
 
 ## [7.1.1] 2024-10-28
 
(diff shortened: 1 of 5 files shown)


davinov (approved):
Indeed, thanks for this fix and the associated test!

toucan_connectors/http_api/pagination_configs.py
```diff
@@ -56,6 +56,13 @@ class OffsetLimitPaginationConfig(PaginationConfig):
     offset: int = Field(0, **UI_HIDDEN)
     limit_name: str = "limit"
     limit: int
+    data_length_filter: str = Field(
+        ".",
+        description=(
+            "Filter to determine the data length and compare it to the limit value. "
+            "It must point to a list of results. " + FilterSchemaDescription
```
lukapeschke, line 63:
```suggestion
            "Filter to access the received data. Allows to compare its length to the limit value. "
            "It must point to a list of results. " + FilterSchemaDescription
```

Wouldn't `data_filter` be a better name ?

CHANGELOG.md
```diff
@@ -2,6 +2,11 @@
 
 ## Unreleased
 
+### Fix
+
+- HTTP API: Extract API results dataframes separately and merge them after.
+- HTTP API: Add `data_length_filter` offset pagination config field to determine which part of data must be used to compute the data length.
```
lukapeschke, line 8: Shouldn't this be in the `Added` section ?

CHANGELOG.md
```diff
@@ -2,6 +2,11 @@
 
 ## Unreleased
 
+### Fix
+
+- HTTP API: Extract API results dataframes separately and merge them after.
```
lukapeschke, line 7: The changelog is destined to end-users. Instead of explaining the code change, could you explain the issue that was solved ?

lukapeschke (changes requested):
Thanks! few nits here and there. And `pandas.concat` should be compatible with v1 and v2, AND do what you want :slightly_smiling_face:

lukapeschke (approved):
thx!

PR #1818, opened 2024-11-13
feat: support python 3.13

lukapeschke:


lukapeschke:
GBQ and protobuf still broken, snowflake does not publish py3.13 wheels yet (see https://github.com/snowflakedb/snowflake-connector-python/issues/2087), and seems like there's a missing `cgi` module with athena

lukapeschke:
Bumper `proto-plus` after reading https://github.com/googleapis/python-bigquery-storage/issues/842 , https://stackoverflow.com/questions/79103765/typeerror-couldnt-build-proto-file-into-descriptor-pool-with-google-cloud-co, https://github.com/protocolbuffers/protobuf/issues/18706

Fanaen (approved):
Thanks 👍

julien-pinchelimouroux (approved):

PR #1825, opened 2024-11-19
fix(mongo): igore index when concatenating chunks

fspot:
https://github.com/ToucanToco/toucan-connectors/pull/1813 introduced an error because each chunk has an index starting at 0 so further transformation lead to
```
ValueError: cannot reindex on an axis with duplicate labels
```
and weird behaviour like 
image
:upside_down_face:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,12 +2,15 @@
 
 ## Unreleased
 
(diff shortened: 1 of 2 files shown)


lukapeschke (approved):
good catch!

PR #1848, opened 2024-12-06
fix(tests): wait for CI server to be ready in azure connector fixture

lukapeschke:


Fanaen (approved):
Looks good to me! 
Not really up to speed with pytest so another review would be appreciated!

julien-pinchelimouroux (approved):

PR #1849, opened 2024-12-11
fix(common): prevent injection during jinja templating (v2.23)

davinov:
Security issue: some users could inject malicious code in jinja templates in queries if we don't use the sandboxed environment


lukapeschke (approved):
Thanks!

PR #1850, opened 2024-12-11
fix(common): prevent injection during jinja templating

davinov:
Forward port of #1849

### DIFF
--- a/toucan_connectors/common.py
+++ b/toucan_connectors/common.py
@@ -23,8 +23,6 @@ if TYPE_CHECKING:  # pragma: no cover
 class NativeImmutableSandboxedEnvironment(NativeEnvironment, ImmutableSandboxedEnvironment): ...
 
 
-jinja_env = ImmutableSandboxedEnvironment()
-
 # Query interpolation
 
 RE_PARAM = r"%\(([^(%\()]*)\)s"
@@ -173,7 +171,7 @@ def _render_query(query: dict | list[dict] | tuple | str, parameters: dict | Non
             clean_p = _prepare_parameters(clean_p)  # type:ignore[assignment]
             env: Environment | NativeEnvironment = NativeImmutableSandboxedEnvironment()
         else:
-            env = jinja_env  # noqa: S701
+            env = ImmutableSandboxedEnvironment()  # noqa: S701
 
         try:
             res = env.from_string(query).render(clean_p)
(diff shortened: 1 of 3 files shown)


toucan_connectors/common.py
```diff
@@ -19,6 +20,11 @@
     import sqlalchemy as sa
 
 
+class NativeImmutableSandboxedEnvironment(NativeEnvironment, ImmutableSandboxedEnvironment): ...
+
+
+jinja_env = ImmutableSandboxedEnvironment()
+
```
lukapeschke, line 27:
```suggestion
```

toucan_connectors/common.py
```diff
@@ -165,9 +171,9 @@ def _render_query(query: dict | list[dict] | tuple | str, parameters: dict | Non
 
         if is_jinja_alone(query):
             clean_p = _prepare_parameters(clean_p)  # type:ignore[assignment]
-            env: Environment | NativeEnvironment = NativeEnvironment()
+            env: Environment | NativeEnvironment = NativeImmutableSandboxedEnvironment()
         else:
-            env = Environment()  # noqa: S701
+            env = jinja_env  # noqa: S701
```
lukapeschke, line 176:
```suggestion
            env = ImmutableSandboxedEnvironment()
```

toucan_connectors/common.py
```diff
@@ -237,7 +243,7 @@ def _flatten_dict(p, parent_key=""):
         parameters.update(p_keep_type)
 
     logging.getLogger(__name__).debug(f"Render query: {query} with parameters {parameters}")
-    return Template(query).render(parameters)
+    return jinja_env.from_string(query).render(parameters)
```
lukapeschke, line 246: Same here, I'm in favor of avoiding globals when possible

toucan_connectors/common.py
```diff
@@ -476,7 +482,7 @@ def pandas_read_sql(
     if convert_to_printf:
         query = convert_to_printf_templating_style(query)
     if render_user:
-        query = Template(query).render({"user": params.get("user", {})})
+        query = jinja_env.from_string(query).render({"user": params.get("user", {})})
```
lukapeschke, line 485: same

lukapeschke (changes requested):
Thanks! Would be great if we could avoid the global var

davinov:
> Thanks! Would be great if we could avoid the global var

Indeed, instantiating an env is cheap, so no need for a global:
[d36ce8f](https://github.com/ToucanToco/toucan-connectors/pull/1850/commits/d36ce8f05d5a9940a14a35b87f6bab89f7e9f878)

lukapeschke (approved):
Thanks!

PR #1883, opened 2025-01-21
feat: switch from poetry to uv

lukapeschke:


lukapeschke:
Somehow this also triggerred a long due `ruff format` :sweat_smile:

Fanaen (approved):
That's great! 👍 
\<insert pun on UVs here>

PR #1891, opened 2025-01-29
feat: accept null refresh-token on oauth-tokens

julien-pinchelimouroux:


toucan_connectors/http_api/authentication_configs.py
```diff
@@ -85,8 +85,11 @@ class OauthTokenSecretData(BaseModel):
     model_config = ConfigDict(populate_by_name=True)
 
     access_token: str = Field(..., alias="accessToken")
-    refresh_token: str = Field(..., alias="refreshToken")
-    expires_at: float = Field(..., alias="expiresAt")  # timestamp
+    # Can't be None because the secret keeper will set a default value if it is missing
+    expires_at: float  # timestamp
```
lukapeschke, line 89: why did you drop the alias ? Also, what will the default value be if the token has no expiration date ?

toucan_connectors/http_api/authentication_configs.py
```diff
@@ -246,14 +249,17 @@ def _get_access_token(self) -> str:
...
-            new_token = client.refresh_token(self.token_url, refresh_token=oauth_token.refresh_token)
-            oauth_token = self._secrets_keeper.save(
-                self.auth_flow_id,
-                # refresh call doesn't always contain the refresh_token
-                new_token | {"refresh_token": oauth_token.refresh_token},
-            )
+        if oauth_token.refresh_token:
+            # If refresh token exists, we want to verify if the access_token is still valid
+            # or if it must be refreshed.
+            if datetime.fromtimestamp(oauth_token.expires_at) < datetime.utcnow():
```
lukapeschke, line 255:
```suggestion
            if datetime.fromtimestamp(oauth_token.expires_at) <= datetime.utcnow():
```

don't we want a little safety margin here ?  Like 30s ?

Fanaen: Earlier or later? 

I'd say a maybe-unnecessary refresh is not a problem, but returning a token really close to expiration may lead to issues down the road (in which case, I second the safety margin).

Fanaen (commented):
I'm a little concerned about the the fact we would use the absence of `refresh_token` to infer we got an unexpirable token. If for some reason, we loose the refresh_token, it would lead to a case where what happened is not clear cut and difficult to debug. 

What do you think about having an enum or a flag UnexpirableToken of something like that?

julien-pinchelimouroux:
@Fanaen I'm not sure how an additional flag on the oauth-token model would help us if the issue is that the refresh-token is missing
If we have something like 
```python

class OauthTokenSecretData
    access_token = "access-123"
    refresh_token = None
    is_infinite_token = False
 ```
 What more do we know ? 
 If that case occurs, we have to check if that oauth2 service is supposed to return a refresh-token, and then check how it can be missing.

toucan_connectors/http_api/authentication_configs.py
julien-pinchelimouroux: [649ff5a](https://github.com/ToucanToco/toucan-connectors/pull/1891/commits/649ff5a55939f8e44fad2ae3813a6567f7ed9c8f)

toucan_connectors/http_api/authentication_configs.py
julien-pinchelimouroux: [649ff5a](https://github.com/ToucanToco/toucan-connectors/pull/1891/commits/649ff5a55939f8e44fad2ae3813a6567f7ed9c8f)

Fanaen (approved):
Indeed, you're right. The added flag doesn't tell much. 
Thanks for the fix 👍

PR #1950, opened 2025-04-01
fix: raise an error if selected column names are parameterized (TCTC-10369)

julien-pinchelimouroux:


toucan_connectors/common.py
```diff
@@ -496,6 +500,8 @@ def pandas_read_sql(
         query = query.replace("%%", "%")
         query = re.sub(r"%[^(%]", r"%\g<0>", query)
         df = pd.read_sql(query, con=con, params=params, **kwargs)
+        if "?column?" in df.columns:
+            raise SelectedColumnInterpolationError("Selected column names cannot be parametrized.")
```
lukapeschke, line 504: Seems dangerous to me: the query has already been executed. Also, the column name might change

lukapeschke (changes requested):
I'm not sure this is how we should proceed, let's discuss it

Fanaen:
Since the topic was discussed and we chose to not fix, we could close this PR, right?

PR #1943, opened 2025-03-28
feat!(mssql): allow custom certificates and use sql alchemy instead [TCTC-10323]

Fanaen:
## Change Summary

- adds `mssql:2022-latest` to docker file and adapts the tests to check on both `2019` and `2022`
- uses `sqlalchemy` instead of `pyodbc` directly.
- adds `trust_server_certificate` to `MSSQLConnector`
- removes `MSSQLConnector.get_connection_params()`

## TODO

- [x] Use MSSQLConnector for AzureMSSQLConnector, under the hood
- [x] adapt tests with pyodbc mocks

## Related issue number

https://toucantoco.atlassian.net/browse/TCTC-10323

## Checklist

* [x] Unit tests for the changes exist
* [x] Tests pass on CI and coverage remains at 100%
* [x] Documentation reflects the changes where applicable

PR #1966, opened 2025-04-11
fix(mssql): restore error handling in get_form

Fanaen:
We removed a `suppress` in https://github.com/ToucanToco/toucan-connectors/pull/1943.
Turns out, it was there to support the case where the user is not allowed to access database schemas.
I put it back with additional tests.

It was the case for our Azure CI, which worked previously since Azure MSSQL did not have this database search. 

Maybe we should add a short timeout to improve UX.

### DIFF
--- a/toucan_connectors/mssql/mssql_connector.py
+++ b/toucan_connectors/mssql/mssql_connector.py
@@ -29,6 +29,8 @@ from toucan_connectors.toucan_connector import (
 if TYPE_CHECKING:
     import sqlalchemy as sa
 
+_LOGGER = getLogger(__name__)
+
 
 class MSSQLDataSource(ToucanDataSource):
     # By default SQL Server selects the database which is set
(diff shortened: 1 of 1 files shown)


lukapeschke (changes requested):
Argh bad review from my side, I let https://github.com/ToucanToco/toucan-connectors/pull/1943#discussion_r2020722275 slide :grimacing: 

Could you please replace the `suppress` with a `try/except` block + a log message in case of an error ?

Fanaen:
Aaah, my bad too. I'm on it!

CHANGELOG.md
```diff
@@ -4,7 +4,9 @@
 
 ### Fixed
 
-- MSSQL: get form even if database search is unsuccessful
+
```
lukapeschke, line 7:
```suggestion
```

toucan_connectors/mssql/mssql_connector.py
```diff
@@ -90,6 +89,8 @@ def get_form(cls, connector: "MSSQLConnector", current_config):
                         constraints["table"] = strlist_to_enum("table", available_tables, None)
 
                     cursor.close()
+        except Exception as exc:
+            getLogger(__name__).warning(f"unable to fetch database and/or table lists for {__name__}: {exc}")
```
lukapeschke, line 93: Please declare a top-level `_LOGGER` variable, and please use `exc_info` for exceptions, it will allow us to have a proper stack trace:

```suggestion
            _LOGGER.warning(f"unable to fetch database and/or table lists for {__name__}", exc_info=exc)
```

Fanaen: [607ba51](https://github.com/ToucanToco/toucan-connectors/pull/1966/commits/607ba5129f8a8edd85ebe6761826559fe26786a8)

lukapeschke (approved):

PR #1967, opened 2025-04-11
fix(azure): revert get_form behaviour

Fanaen:
Azure MSSQL: database search added by accident in get_form, removed it


Fanaen:
We want the database fetch, so this way.

PR #1969, opened 2025-04-14
fix(mssql): only wait 5 seconds for the database and table list in connector form

Fanaen:


Fanaen:
Blocked by https://github.com/ToucanToco/toucan-connectors/pull/1966

lukapeschke (approved):

PR #1978, opened 2025-05-05
test(mssql): discard login errors while waiting for the db to start

Fanaen:
While the MSSQL database is booting up, the login may fail. This was introducing flakyness in the tests cases.

## Change Summary

- discard `InterfaceError` as well as `OperationError`. If login is indeed broken (which should be rare), it will end up in a timeout error instead, which is fine.

## Related

https://github.com/ToucanToco/toucan-connectors/actions/runs/14832515032/job/41636596419

## Checklist

* [ ] ~~Documentation reflects the changes where applicable~~


lukapeschke (approved):
Thanks!

PR #2162, opened 2026-07-08
feat: postgres supports string ports [TCAI-1366]

julien-pinchelimouroux:


Fanaen (approved):

toucan_connectors/postgres/postgresql_connector.py
```diff
@@ -121,7 +121,7 @@ class PostgresConnector(
     """
 
     host: str | None = Field(None, description="The listening address of your database server (IP adress or hostname)")
-    port: int | None = Field(None, description="The listening port of your database server")
+    port: int | str | None = Field(None, description="The listening port of your database server")
```
lukapeschke: I think this should always be an int with a pre-validator

lukapeschke (changes requested):

toucan_connectors/postgres/postgresql_connector.py
Fanaen: Oh, so something like "toto" is not representable state, right? Good idea!

tests/postgres/test_postgres.py
```diff
@@ -272,6 +272,19 @@ def test_no_user():
         PostgresConnector(host="some_host", name="test")
 
 
+def test_bad_string_port():
+    """It should raise an error as the port is not a number"""
```
lukapeschke: when the port is not a number

toucan_connectors/postgres/postgresql_connector.py
```diff
@@ -121,7 +121,9 @@ class PostgresConnector(
     """
 
     host: str | None = Field(None, description="The listening address of your database server (IP adress or hostname)")
-    port: int | str | None = Field(None, description="The listening port of your database server")
+    port: Annotated[int | str | None, BeforeValidator(lambda x: int(x) if isinstance(x, str) else x)] = Field(
```
lukapeschke: Why not `BeforeValidator(int)` ?

lukapeschke: also, typing should be `int | None`

toucan_connectors/postgres/postgresql_connector.py
```diff
@@ -139,10 +141,10 @@ class PostgresConnector(
...
 
     def _get_port(self) -> int | None:
-        port = self.port
-        if isinstance(port, str):
-            port = int(port)
-        return port
+        if self.port and isinstance(self.port, str):
+            return int(self.port)
+        else:
+            return self.port
```
lukapeschke: not needed anymore

lukapeschke (approved):

